PR #82, opened 2015-01-19
Fixed typo in a potentially user-facing error message.

rdbell:


elazarl:
Thanks! Appreciate those fixes. Hard to catch, yet, makes all the difference.

PR #84, opened 2015-01-26
close the target and proxy connections after both finished talking

ArielBadichi:


elazarl:
Hi,

Thanks for the PR.

Can you add a test case explaining the problem?

elazarl:
Hi,

I just want to explain. I've had a similar problem in the past with HTTPS, so I wanted to take the time and make sure it all works.

I understand that waiting is frustrating to the contributor, and indeed I've had little time for goproxy lately, I assure you I'll reach to that soon.

Thanks and sorry,

mattweidner:
Prehaps a better way to handle this would be for client code to se http.Transport variables IdleConnTimeout and ResponseHeaderTimeout. They can be set on a new goproxy ProxyHttpServer instance. The default for those timeouts is 0, which waits forever. Setting it to a non-zero value will automatically close a connection when the timeout lapses.
```
package main

import (
        "log"
        "time"
        "github.com/elazarl/goproxy"
)

func main() {
        var timeout time.Duration = 500 * time.Millisecond
        proxy:= goproxy.NewProxyHttpServer()
        log.Println("Idle timeout: ",proxy.Tr.IdleConnTimeout)
        log.Println("Response header timeout: ",proxy.Tr.ResponseHeaderTimeout)
        proxy.Tr.IdleConnTimeout = timeout
        proxy.Tr.ResponseHeaderTimeout = timeout
        log.Println("Idle timeout: ",proxy.Tr.IdleConnTimeout)
        log.Println("Response header timeout: ",proxy.Tr.ResponseHeaderTimeout)
}
```

mattweidner:
@ArielBadichi Perhaps a better way to handle this would be for your application code to set the http.Transport variables IdleConnTimeout and ResponseHeaderTimeout. They can be set on a new goproxy ProxyHttpServer instance. The default for those timeouts is 0, which waits forever. Setting it to a non-zero value will automatically close a connection when the timeout elapses.
```
package main

import (
        "github.com/elazarl/goproxy"
        "log"
        "time"
)

func main() {
        var timeout time.Duration = 500 * time.Millisecond
        log.Println("Idle timeout: ", proxy.Tr.IdleConnTimeout)
        log.Println("Response header timeout: ", proxy.Tr.ResponseHeaderTimeout)
        proxy.Tr.IdleConnTimeout = timeout
        proxy.Tr.ResponseHeaderTimeout = timeout
        log.Println("Idle timeout: ", proxy.Tr.IdleConnTimeout)
        log.Println("Response header timeout: ", proxy.Tr.ResponseHeaderTimeout)
}

```

PR #85, opened 2015-01-26
close response body after reading it

ArielBadichi:


elazarl:
Hi,

Can you add a test case which is solved by the commit? Would help me understand it.

ArielBadichi:
Added test case.

elazarl:
Thanks, would have a look later today.

elazarl:
Would take care of that shortly, have had busy days.

Sorry.

ErikPelli:
In the current master branch, body is closed with a defer. Closing this PR.

PR #87, opened 2015-01-30
Adding in RemoteAddr field to requests converted from CONNECT

mzimmerman:
If you want to filter proxy requests by who connected, the RemoteAddr field of the *http.Request needs to be populated on the *http.Request object when converting from a CONNECT request


elazarl:
Thanks,

PR #91, opened 2015-02-21
Rename examples

pmezard:
Hello,

Here is a bit of work on the examples and their documentation. The first changeset is a big rename, I prefer to discuss it before moving forward. Tell me if you want it done differently. More will follow once this pull-request is merged.


elazarl:
Hi,

Thanks for the PR.

The naming change looks like a stutter to me.

Why do you need to repeat goproxy in the examples name, it already at `github.com/elazarl/goproxy/example/foo`. Why do you need to have another `goproxy-` prefix?

pmezard:
I suggested it here https://groups.google.com/d/msg/goproxy-dev/1-JW__ZhhFA/7fs2HEoOBzcJ , it is to avoid polluting Go bin directory with random binaries from examples. I can remove it if you are not convinced, that is why I wanted to start with the renaming business.

elazarl:
Hmmm... maybe we should put those directories inside "internal" folder, so that they won't be built?

Can you ask golang-nuts what's the best practice with including examples in your repo, without polluting bin/ namespace? This is a good point (I personally rarely use `...`, but obviously some people do...).

pmezard:
Followup here https://groups.google.com/forum/#!topic/golang-nuts/6JeU_A01vzg . Matthew idea of using build tags sounds smarter than my prefixing.

PR #97, opened 2015-03-19
[RFC] Implement SNI sniffing based on github.com/inconshreveable/go-vhost

abourget:
Changes the behavior slightly for the "ConnectAccept" case, which will
always send a "200 OK", but will shut down the connection if the remote end
doesn't answer.

The SniffSNI property on the ProxyHttpServer is a proposition, so we could
handle both cases.  For those who don't need SNI, we could skip the overhead,
but those who enable it, would accept the altered "ConnectAccept".

The code doesn't deal with this switch, but if the idea is accepted, I
could rework it to make it a nicely switchable feature.


elazarl:
Hi,

Thanks for the PR.

I don't understand, why won't you sniff the host only when ConnectMitm was requested? That is, in `case ConnectHTTPMitm`?

The current code breaks all other behavior for connect request.

PS,
The correct name for the function is `handleConnect`. Putting HTTPS there was probably my mistake.

abourget:
The reason is that when you loop the `proxy.httpsHandlers`, you need the real host at that point.. so you can take your decision whether you Accept, or Mitm or Hijack, etc.. otherwise, all you have is the CONNECT hostname.. and this precisely is why we implement SNI :) because CONNECT isn't sufficient.

For example, the Genymotion Android emulator has a way to proxy any outgoing connections through a proxy, but does the resolution, so always sends a CONNECT with the IP address, you then need to sniff the SNI to get the hostname the android machine is actually asking, and expects in the Server cert CommonName..

The PR code only changes the behavior `ConnectAccept` .. all other sent the 200 OK anyway... and the vhost "restores" the `net.Conn` to what it was before sniffing.. so nothing else should change, except you have the ServerName requested through SNI.

I agree with the function name.

abourget:
I think we could make it configurable, at the "Proxy" level, that we do want to enable Sniffing.. as it changes the behavior for ConnectAccept, and the `host` used is not the same.

I was thinking of other refactorings also, like having an `UpstreamRequest` and `DownstreamRequest`, `UpstreamResponse` and `DownstreamResponse`.. the `Upstream*` we read, and `Downstream*` we write or modify. Also, having the `ProxyCtx` be used a bit more, where the `Host`, or `ConnectHost` or `SNIHost` could be read, and decisions could be taken based on some combinations of those, etc..  having `host` as a first-level argument makes it difficult to pass more information about where we got that `host`.. having that in the `ProxyCtx` would allow to add more things.

I was thinking of these things because I'll need to implement the equivalent of "Map Local" and "Map Remote" options in Charles Proxy.. and doing so with the current design is a bit hacky, and quite verbose.

elazarl:
@abourget what I suggest is, add a ConnectAccept mode that would sniff TLS, and would be able to accept/reject/MITM TLS request according to the SNI field.

What do you think?

I think that the user would usually want to deal with the SNIHost, so if anything, we need to add an ConnectHost field that would be populated only for CONNECT requests.

What are the "map local" and "map remote" options in Charles Proxy?

abourget:
I can see. but this makes the usage verryy clunky and difficult to reason about. Think about this:

```
proxy.OnRequest(goproxy.DstHostIs("my.host.com")).HandleConnectFunc(func(host string, ctx *goproxy.ProxyCtx) (*goproxy.ConnectAction, string) {
		return &goproxy.ConnectAction{Action: goproxy.ConnectTHROUGHSNISNIFFING, TLSConfig: goproxy.TLSConfigFromCA(&myOwnGoproxyCa)}, host
	})
```

you're only asking after `HandleConnectFunc` how to sniff, but following the chaining model of `OnRequest()` and `HandleConnectFunc()`.. you would expect to already know it before triggering this handler (because you specified it first in the chaining)..  What would `host` be in the first param to the `HandleConnectFunc` ? what should we return as a `host` in there, since we're only then *asking* for SNI sniffing ?

I feel the whole chaining model is a little weird.. and the HandleConnect, and ConnectAction, `*Connect`, containing `ConnectActionLiteral`... I'm trying to think of a simplified model, with less `Connect` keywords :) this term feels really overloaded in this context.

elazarl:
The thing is, there are two separate things we want to let the user do:

1) Handle CONNECT requests without caring about the content. Useful for
"regular" transparent proxy.
2) MITM CONNECT requests, based on CONNECT host.
3) Assume a CONNECT request is HTTPS, peek the SNI, and then decide what to
do based on the SNI host.

The domain is complex, and I want the user to have the flexibility to both
use, or ignore the SNI.

On Mon, Mar 23, 2015 at 3:55 PM, Alexandre Bourget <notifications@github.com

abourget:
Digression follows:

The only thing I can see that can be done, to discriminate if we should do SNI, is based on the CONNECT hostname.. but the current `OnRequest()` model doesn't allow for that.  We'd need a more fully fledged context, and have fewer methods, that leave more room for userland filtering and decision. Maybe a model where you simply get contexts, take decisions, and return contexts with its content determining what happens next ?

Maybe having merely an `OnRequest(func(ctx *ProxyCtx) *ProxyCtx {})` ... (and similarly for `OnResponse()`), and documenting what the returned ProxyCtx expects and how it will take action based on that return value.  This would put more logic on the user side, but maybe we could add more logic to the ProxyCtx, like helpers.

elazarl:
The connect action can make a decisions based on the CONNECT hostname.

OnRequest is relevant only if you decide to MITM. It doesn't make sense
otherwise. You can't see the request without MITM'ing.

I'll be happy to see usage examples of what you've had in mind.

Most of the work here is a good design, and I'm grateful you're helping out.

On Mon, Mar 23, 2015 at 4:00 PM, Alexandre Bourget <notifications@github.com

abourget:
> OnRequest is relevant only if you decide to MITM. It doesn't make sense
otherwise. You can't see the request without MITM'ing.

I see.. but the OnRequest() precedes the handlers in the `.` chain.. isn't that backwards ?

abourget:
Are you available to https://gitter.im/elazarl/goproxy today ?

abourget:
I think we can close this.. my code base has diverged so heavily that it's kind of a hard-fork.

I've moved it to github.com/abourget/goproxy .. I might rename it if you don't want the association or the confusion.

elazarl:
Thanks,

I was thinking of a breaking API change to solve that.

Would you consider merging the two projects when this is done?

abourget:
Oh definitely.  It'd be way better merged back. But it's quite a departure
from what was there before.. and since I got no feedback.. I thought it
wasn't the way you guys wanted to go.

Would you like to have a little chat session to discuss this ?

elazarl:
I'd love to. My gmail handle is the same as the github handle.

abourget:
I contacted you through Hangouts. Does that work for you ?

pytlesk4:
Did you guys ever get a chance to do a hangout?  Wondering if these two projects are going to be merged.

abourget:
Anything you'd like to see merged ?

elazarl:
Sorry, not yet. Will try to think it over next week.

pytlesk4:
@abourget I switched to your forked project, so would preferably like to see most of it merged.  I like the interface of the dispatcher a lot.

abourget:
@taliesins I see you've written a lot of `goproxy` code.. I was wondering if you took a look at my fork, and what you thought about it.  We were discussing eventually merging it as goproxy v2 or something.. your experience would be valuable as input. see it here: https://github.com/abourget/goproxy

elazarl:
PS,

I apologize for the V2 design not happening yet.

Many personal issues lately.

I *do* want to make it in the near future.

Stay tuned.

On Fri, Jan 29, 2016 at 6:49 AM, Alexandre Bourget <notifications@github.com

PR #99, opened 2015-04-14
More examples documentation

pmezard:
Hello,

Here is a little more documentation and cleanup of goproxy examples. It is far from done but this might help newcomers a bit. It certainly improved my understanding of goproxy by reading and tidying up some code. This is a cosmetic pass, some examples seem too complicated like the one logging HTTP content: the logging code, while interesting, seems a bit irrelevant and makes things harder by adding a lot of logic for clean process termination. I may go back to that once I have a clear picture of the remaining samples.

Also, what about gofmt'ing the code?

### DIFF
--- a/dispatcher.go
+++ b/dispatcher.go
@@ -43,7 +43,7 @@ func (c RespConditionFunc) HandleResp(resp *http.Response, ctx *ProxyCtx) bool {
 	return c(resp, ctx)
 }
 
-// UrlHasPrefix returns a ReqCondtion checking wether the destination URL the proxy client has requested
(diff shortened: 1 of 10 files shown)


elazarl:
Regarding gofmt, generally speaking the code should be gofmt'd, but I probably missed something. I think the best approach to fix it, is to add precommit git hooks, and to fix everything in a single commit.

elazarl:
@pmezard 

Improving documentation and fixing typos is a _very_ important job, with a very little glamour.

I applaud your efforts.

Thanks!

Issue #81, opened 2015-01-18
Tiny error in error return in example httpdump.go


pschlump:
In httpdump.go example around line 190 you have

func (t *TeeReadCloser) Close() error {
	err1 := t.c.Close()
	err2 := t.w.Close()
	if err1 == nil && err2 == nil {
		return nil
	}
	if err1 != nil {
		return err2 /// This always return NIL
	}
	return err1
}

This should be:

func (t *TeeReadCloser) Close() error {
	err1 := t.c.Close()
	err2 := t.w.Close()
	if err1 == nil && err2 == nil {
		return nil
	}
	if err1 != nil {
		return err1 // Changed err2 to err1
	}
	return err2 // Changed err1 to err2
}

OR:

func (t *TeeReadCloser) Close() error {
	err1 := t.c.Close()
	err2 := t.w.Close()
	if err1 == nil && err2 == nil {
		return nil
	}
	if err1 == nil {  // Changed != to ==
		return err2 
	}
	return err1
}

elazarl:
Thanks.

Can you send a pull request to fix that?

pschlump:
I can send a pull request.  It will be a few days.
I would like to have a test to validate this - but the test is much more
complicated than the fix and I have to get a better understanding of the
code.

I am also getting this error and looking for the cause.  This happens after
it has been idle for a long period.

goroutine 906 [IO wait, 31 minutes]:
net.runtime_pollWait(0x5b1750, 0x72, 0x0)
/private/var/folders/00/0sdwh000h01000cxqpysvccm0035qk/T/makerelease003761064/go/src/pkg/runtime/netpoll.goc:146
+0x66
net.(*pollDesc).Wait(0xc208298b50, 0x72, 0x0, 0x0)
/usr/local/go/src/pkg/net/fd_poll_runtime.go:84 +0x46
net.(*pollDesc).WaitRead(0xc208298b50, 0x0, 0x0)
/usr/local/go/src/pkg/net/fd_poll_runtime.go:89 +0x42
net.(*netFD).Read(0xc208298af0, 0xc2083dd000, 0x1000, 0x1000, 0x0,
0x5a93c8, 0x23)
/usr/local/go/src/pkg/net/fd_unix.go:242 +0x34c
net.(*conn).Read(0xc20802e950, 0xc2083dd000, 0x1000, 0x1000, 0x0, 0x0, 0x0)
/usr/local/go/src/pkg/net/net.go:122 +0xe7
[...]

PR #101, opened 2015-04-19
httpdump: make TeeReadCloser.Close fail if either output fails

pmezard:
This should fix https://github.com/elazarl/goproxy/issues/81

I did not touch it when refactoring the code because the intent was not clear to me. But according to the issue, you want to fail fast, which makes sense, then just let do that.


elazarl:
Pretty obviously a bug.

Thanks @pmezard , you're pretty active lately, I hardly catch up ;-)

PR #111, opened 2015-06-22
added support for authenticated proxies

jdamick:
set the authorization header for proxies with authentication using tunneling


elazarl:
Hi,

Did you see the [ext/auth](https://github.com/elazarl/goproxy/tree/master/ext/auth) library?

Doesn't it achieve the same?

jdamick:
unless i'm misunderstanding, that is for authenticating to goproxy, not authenticating to an upstream proxy since it doesn't add the proxy-authorization header.. ?

elazarl:
In that case, simply use authinticated transport. Add it to the context.

jdamick:
i can do as you suggest but this change would make the default https transport match the default http transport functionality as that doesn't require anything special for setting the authentication headers since the underlying http.Transport does this in the getConn method which is similar to the https.NewConnectDialToProxy..

elazarl:
Hey,

Can you please add it as `ext/` extension, and  not in core package?

Thanks

ErikPelli:
As already stated, proxy authentication can be set in the proxy URL, starting from the cascadeproxy example.
This pull request wouldn't add anything that can be useful for the core library, closing it.

PR #114, opened 2015-06-25
Context issue when HTTPS is used in MITM mode

g3rk6:
Fixed the issue which goproxy doesn't provide correct HTTP request information from the context when HTTPS MITM is used.


elazarl:
@g3rk6 Thanks! Great catch, since I usually use the request function parameter, I didn't check the context.

PR #115, opened 2015-06-27
Fixed the issue which response error code and body were never submitted to response filters.

g3rk6:
Running response filters on 'nil' response was problematic and not very helpful. I.e. if response filter function uses channel, etc. Therefore, it limits the possibility of response filter doing things based on status code 500 or body which contains the error.

### DIFF
--- a/proxy.go
+++ b/proxy.go
@@ -111,15 +111,11 @@ func (proxy *ProxyHttpServer) ServeHTTP(w http.ResponseWriter, r *http.Request) 
 			removeProxyHeaders(ctx, r)
 			resp, err = ctx.RoundTrip(r)
 			if err != nil {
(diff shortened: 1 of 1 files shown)


g3rk6:
Any thoughts? Or should I close the pull request? Thanks.

elazarl:
Hi,

I apologize for not responding. I appreciate every pull request and yours somehow slipped between the chairs.

My problem with your commit, is, it prevents the user from sending a custom error message if the request was unsuccessful.

For example, if someone wants to first try and fetch content from the web, and if that not available, use his own internal cache, this would be impossible with your commit.

Moreover, one can implement your commit by a simple response filter.

So I'm not sure if users would benefit from this change.

What do you think?

g3rk6:
Ah, that makes sense and thanks for the tips. I'm going to close this one then!

PR #116, opened 2015-07-16
Caching proxy example

taliesins:
Added an example project for a caching proxy server. Example is great for apt-get and windows update caching proxy for packer.


elazarl:
Thanks, looks great.

But wouldn't it better to have it in its own repository? It looks pretty sophisticated.

taliesins:
Its relatively simple and defiantly not full featured. I hacked it together to support Packer. So I am hoping someone will use it for inspiration to create a proper implementation. 

I think this is a common use case as well and I would love to see it included as part of the examples. Good projects like yours need more contributers.

PR #121, opened 2015-09-22
Remove extra slash from UrlHasPrefix

KevinMGranger:
URL.Path already has a leading slash, so adding our own slash causes strings.HasPrefix to fail.


elazarl:
Thanks!

We need to add a test for UrlHasPrefix...

PR #123, opened 2015-10-03
README: link to godoc.org

pmezard:
Because it is really convenient.


elazarl:
@pmezard Thanks! I'll get to the other PR shortly.

PR #127, opened 2015-10-13
Avoid tooo many warns due to clients' closed connections

gallir:
In https, the clients close the connections when they receive the expected data, it's harmless. This patch  avoids warnings if it could write bytes to the clients.


elazarl:
Hi,

Thanks for the PR.

I'm not sure I follow. Why would bytes be <0? Why is that a sign we should ignore the error.

In principle I think removing those warnings is a very good idea.

elazarl:
@gallir ping. Can you please have a look and explain?

ErikPelli:
Fixed by https://github.com/elazarl/goproxy/pull/582

PR #133, opened 2015-11-04
doc(readme): make basic example fully compilable

JensRantil:
Nicer for Golang newbies.


elazarl:
Thanks. Totally worth it.

PR #135, opened 2015-11-25
https: add support for ephemeral certificate generation (w/ ECDSA)

tmthrgd:
An unexported `signEphemeralHost` function produces a self-signed certificate for the hosts. It will use either P-256 if `ecdsaKey` is true or RSA with a 1024bit key otherwise.

An exported `TLSConfigEphemeral` function provides a `tls.Config` struct that will produce an ephemeral certificate during the tls handshake. It prefers ECDSA keys but will fall back to RSA if no ECDSA cipher suites are supported by the client. One generated the key is cached by placing it into the `Config.Certificates` array.

Each of `OkConnect`, `MitmConnect`, `HTTPMitmConnect`, `RejectConnect` has a corresponding ephemeral version `*Ephemeral`.

ECDSA keys are significantly faster to generate and sign with. This reduces the overhead of proxying https requests.

Similar functionality *could* be added to the existing `TLSConfigFromCA` and `signHost` functions but doing so would break compatibility.

### DIFF
--- a/dispatcher.go
+++ b/dispatcher.go
@@ -307,6 +307,21 @@ var AlwaysReject FuncHttpsHandler = func(host string, ctx *ProxyCtx) (*ConnectAc
 	return RejectConnect, host
 }
 
+// AlwaysMitmEphemeral is a HttpsHandler that always eavesdrop https connections, for example to
+// eavesdrop all https connections to www.google.com, we can use
+//	proxy.OnRequest(goproxy.ReqHostIs("www.google.com")).HandleConnect(goproxy.AlwaysMitmEphemeral)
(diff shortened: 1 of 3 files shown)


elazarl:
Hi,

Thanks for the patch.

I don't understand the goal.

Do you want to support ecdsa certificates, since they're faster, or do you want to generate a new CA for every TLS connection request?

tmthrgd:
I was trying to do the former but I've found a much better way.

(Thanks for the great library.)

elazarl:
Thanks.

Can you share this way with us? For other people trying to do the same?

PR #141, opened 2016-01-19
Added CookieMatches for Cookie matching on requests

BBerastegui:
With CookieMatches you will be able to trigger any behaviour if you find regexp matches against some particular cookie pattern.

### DIFF
--- a/dispatcher.go
+++ b/dispatcher.go
@@ -2,6 +2,7 @@ package goproxy
 
 import (
 	"bytes"
+	"fmt"
 	"io/ioutil"
 	"net"
 	"net/http"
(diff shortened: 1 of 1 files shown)


elazarl:
Hi,

Thanks for the PR.

I'm not sure if matching against the cookie representation is a good idea. I'd add a way to match
against a certain cookie's key. So that we'd match a regexp against specific `Cookie(key).Value`.

BBerastegui:
Hi !

In my case, I was trying to find a regexp matching cookie being sent in the request and then modifying/deleting/triggering whatever behaviour we wanted.

That is why I decided to add the same functionality as:
> func ReqHostMatches(regexps ...*regexp.Regexp) ReqConditionFunc

We can add also something more strict as you propose. Something like "CookieIs" as in the host-related function, but based on the cookie key and/or value:
> func ReqHostIs(hosts ...string) ReqConditionFunc

elazarl:
Didn't you know in advance which key would the cookie contain?

On Tue, Jan 19, 2016 at 10:29 AM, Borja Berastegui <notifications@github.com

BBerastegui:
In my specific case yes, one key in particular.

But in the case that the key might be variable (I've seen several applications doing this), it will be great to be able to match it using a regexp.

elazarl:
In this case we should give two regexp, for the key and for the value.

This would avoid all the key escaping problems.

On Tue, Jan 19, 2016 at 10:42 AM, Borja Berastegui <notifications@github.com

BBerastegui:
So, two separate functions like:
> func CookieKeyMatches(regexps ...*regexp.Regexp) ReqConditionFunc

And:

> func CookieValueMatches(regexps ...*regexp.Regexp) ReqConditionFunc

Or one function with two parameters:
> func CookieMatches(regexpsKey ...*regexp.Regexp, regexpsValue ...*regexp.Regexp) ReqConditionFunc

?

PR #143, opened 2016-03-11
Add reverse proxy capability

Quentin-M:
Because why should we limit ourselves to a forward proxy server when we could do both reverse and forward at the same time?

This commit allows us to specify a `ProxyHttpServer` in `NewProxyHttpServer ` that will act as a reverse proxy when the forward proxy constructed will be called as a reverse proxy. When no reverse proxy is specified and the forward proxy is called as a reverse proxy, a response with status code 500 and message `This is a proxy server. Does not respond to non-proxy requests.`, just like before.


elazarl:
Hi,

What would be the advantage to the users for using goproxy as a reverse proxy? There are many other solutions which are probably superior.

I'd like to understand the use case before taking a big change.

Thanks!

Quentin-M:
It allows people who need both a forward and a reverse proxy to use a single library. They can then use the same interfaces for both cases (i.e. enjoying `OnRequest()`).

In our use case, we can have a single server (listening on only one port) doing either request authorization/signing (ie. HMAC or JWT) or request verification depending on how it's called (as a proxy or as a reverse proxy).

elazarl:
Sounds reasonable.

BTW, I think a better approach is, not to add `isReverseProxy`, but instead allow filtering regular HTTP request with the onRequest mechanism.

What do you think?

Quentin-M:
Hi,

Thanks for your response.
However, I am not sure I understand what you just said?

Quentin-M:
Hi,

I noticed that we shouldn't have both the proxy and the reverse proxy on the same port, even though the idea looked nice. Some people might want to terminate TLS on the reverse proxy, however it is impossible to get a forward proxy working properly with TLS termination.

Thus, I modified and simplified slightly my PR. It now simply exposes a new constructor `NewReverseProxyHttpServer()` that creates a reverse proxy. The `NonproxyHandler` will be called when a forward proxy is called like a reverse proxy, and when a reverse proxy is called like a forward proxy.

Quentin-M:
Bump

smintz:
Hey, I was looking for same functionality and bumped into this PR, is it going to be merged?

Quentin-M:
Hi,

Note that in the meantime, you can use https://github.com/coreos/goproxy

smintz:
cool, thanks @Quentin-M!

PR #144, opened 2016-03-15
Set content-length for clients, who don't understand chunked

makhov:
For example, Android emulator can't work with `Transfer-Encoding: chunked` and it's very annoying to me.

### DIFF
--- a/proxy.go
+++ b/proxy.go
@@ -3,11 +3,13 @@ package goproxy
 import (
 	"bufio"
 	"io"
+	"io/ioutil"
 	"log"
 	"net"
 	"net/http"
 	"os"
 	"regexp"
+	"strconv"
 	"sync/atomic"
 )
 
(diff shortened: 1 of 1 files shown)


elazarl:
Hi,

Thanks for the PR.

This is problematic for big requests that won't fit the memory.

Do you have a suggestion what to do?

Maybe we can write a RequestHandler for Android that does just that?

(Apologies for the delay)

2tvenom:
@makhov thank you!
this is work with android devices

ErikPelli:
Is this still required?

ErikPelli:
Maybe could you just follow this discussion to handle the chunked response directly from the Android client? https://stackoverflow.com/questions/48298206/okhttp-retrofit-body-request-transfer-encoding-chunked-added

ErikPelli:
According to this [issue](https://github.com/square/retrofit/issues/2632), okhttp/retrofit will automaticaly unchunk the data, however I can't confirm it.
If this is not the case, reading the whole body is still the wrong approach, that will lead to a much higher (and unnecessary) memory consumption.
The unchunk part should be implemented in another net.http middleware that is outside of the scope of this project, if you really need it, or maybe you can add something on the client side.
Closing this pull request for all these reasons.

Issue #134, opened 2015-11-24
go-charset (no license) switching to fork?


onlyjob:
Apparently _goproxy_ depends on _go-charset_ which is non-free due to [lack of license and copyright statement](https://github.com/rogpeppe/go-charset/issues/1).

Perhaps it might be better to switch to 
https://github.com/paulrosania/go-charset

The latter appears to be better maintained and up to date.

elazarl:
Good idea. Looks like this repository is the one which is maintained.

elazarl:
@onlyjob do you want to submit a PR?

onlyjob:
No, sorry... Too little time and heaps of work to do...

ErikPelli:
https://github.com/elazarl/goproxy/issues/575

PR #145, opened 2016-03-22
Switch to better maintained go-charset fork (Closes: #134).

onlyjob:
When this is merged it would be great to tag new formal release. Thanks.

### DIFF
--- a/ext/html/html.go
+++ b/ext/html/html.go
@@ -9,8 +9,8 @@ import (
 	"net/http"
 	"strings"
 
-	"github.com/rogpeppe/go-charset/charset"
-	_ "github.com/rogpeppe/go-charset/data"
+	"github.com/paulrosania/go-charset/charset"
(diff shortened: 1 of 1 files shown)


elazarl:
That would be great idea.

But I've already did that once, from the google code to rog pepe.

How can we find a repository that would be maintained in the long run?

onlyjob:
Well, google code is gone and at the moment I suggest we switch to project that is currently best maintained. I doubt you can find guarantees that it will stay like this forever. One simply have to adapt to changes. Let's hope that new _go-charset_ proves itself to be the best. :)

onlyjob:
Also worth noticing that nobody can use rogpepe/go-charset due to licensing problems so there is a bigger problem than poorly maintained dependency...

ErikPelli:
I completely removed this dependency in https://github.com/elazarl/goproxy/pull/580. Closing this PR.

PR #153, opened 2016-04-05
Add customizable HTTPS proxy authentication

kgraney:
It is currently impossible to authenticate with a proxy using
Proxy-Authenticate/Proxy-Authorization headers on a CONNECT method.
This adds support by allowing the user to specify a Hijack method that
provides the challenge header.

### DIFF
--- a/https.go
+++ b/https.go
@@ -23,6 +23,7 @@ const (
 	ConnectMitm
 	ConnectHijack
 	ConnectHTTPMitm
+	ConnectProxyAuthHijack
 )
 
 var (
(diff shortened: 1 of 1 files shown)


elazarl:
Sounds reasonable.

Do you think that we should instead allow a mode that controls the response altogether?

I'll merge it, and if another plausible response is due, we'll add a generic mode.

kgraney:
Thanks!  I think the IETF standards only allow a couple different nominal proxy responses.

elazarl:
Care to collect them, we need to add all of them I think.

kgraney:
Well, for authentication they're listed here: https://tools.ietf.org/html/rfc7235#page-7.  For CONNECT specifically, the details are here https://tools.ietf.org/html/rfc2817#page-7.

elazarl:
I was hoping you'll extend the PR for all of them if you have time ;-)

PR #156, opened 2016-04-22
not-ssl websockets will be now served correctly and wont be MITMable

bennyscetbun:
Hi,
We had a problem with http websockets :)
we've made some code to take care of it :)

I hope, you will be ok with it :)


elazarl:
Hi,

Sorry for the late response.

I need to think what is the best way to support websockets.

How does fiddler do it?

bennyscetbun:
Hi Elazarl. I do not know about fiddler.
What i know is that mitmproxy is not able to do websocket at all and that s why we ve been looking at your code. :)
Strangely your code should be able to manage HTTPS websockets :).

elazarl:
but do you know the design decision behind it?

I'd rather find out potential problems with the implementation before
merging it.

bennyscetbun:
Sorry Elazarl, i really dont know what design decision was behind it.
My pull request is a quick fix to enable HTTP Websocket.

Why i have done that is because you remove some http headers that are needed for websockets and you do the roundtrip yourself (which, if i ve understood right, doesnt let the tcp connection between the author and the target opened).

I didnt wanted to change a lot of your code but from a design perspective i would have not done it this way.
I would have done Hijack connection for both HTTP and HTTPS. (right now it s only https :( )
Hijacking connection for HTTP would have let anyone to do their own websocket handler with your library :)

Does it help you ?

bennyscetbun:
Any news ? :)

MeGaPk:
+1, what about that merge or implement same functional?

elazarl:
I really want to support websockets, but I'm a bit worried from the given implementation. I'll try to look into it.

bennyscetbun:
Yep Elazarl. This was a quick fix. you may want to recode it your way.
But for info we are using this code since april. With it we are able to test our app on preprod or locally. Everything is fine :)

MeGaPk:
@elazarl this implementation work with socket.io java version, already tested.

@bennyscetbun need implement onRequest too, for change host, for example.

bennyscetbun:
Hi all, i havent tested his work yet but #187 seems a better work than my small quick dirty fix :)
You should look at it @MeGaPk

elazarl:
@bennyscetbun if you want to tidy this PR, I'll be happy to merge it. Unfortunately I do not have time to do this myself...

PR #161, opened 2016-05-17
Improve copyAndClose implementation for https connections

ap4y:
- Use half-close methods during copy
- Wait for the copy to be finished by waiting on a channel

I had a lot of `Error copying` errors on our servers for the https connections, in some cases client applications were failing with timeouts. At the moment second copy operation is failing because required half of the connection is closed by the first copy. It also worth waiting for the both copy operations to finish. Similar issue was recently fixed in go-socks5, in [this](https://github.com/armon/go-socks5/commit/66ca0497dfb58c4daaf663f53fc8bef7ded309be#diff-d9c546c27e20f6a2acfc4ac7b74c1a26) commit.

### DIFF
--- a/https.go
+++ b/https.go
@@ -98,8 +98,11 @@ func (proxy *ProxyHttpServer) handleHttps(w http.ResponseWriter, r *http.Request
 		}
 		ctx.Logf("Accepting CONNECT to %s", host)
 		proxyClient.Write([]byte("HTTP/1.0 200 OK\r\n\r\n"))
-		go copyAndClose(ctx, targetSiteCon, proxyClient)
-		go copyAndClose(ctx, proxyClient, targetSiteCon)
+
+		tTargetSiteCon := targetSiteCon.(*net.TCPConn)
+		tProxyClient := proxyClient.(*net.TCPConn)
+		go copyAndClose(ctx, tTargetSiteCon, tProxyClient)
+		go copyAndClose(ctx, tProxyClient, tTargetSiteCon)
 	case ConnectHijack:
 		ctx.Logf("Hijacking CONNECT to %s", host)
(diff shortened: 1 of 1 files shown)


elazarl:
Looks like a good idea, but why wait on the error channel?

Logging is synchronized anyway, and we don't care which thread would it
happen at.

ap4y:
Sorry it wasn't clear from my description. I added waiting not for the logging purposes, but to make code of `handleHttps` more deterministic and also unify behavior across switch shoulders. I think it will be easier to reason about this method if connections will be always fully handled when it returned, rather than still being processed in one of the goroutines. I don't have a strong feelings about that part, I can remove it if you feel it's unnecessary.

elazarl:
@ap4y another thing if possible. I think it's better to have copyAndClose accept a TCP connection, since it makes little sense otherwise, you need WriteClose in order for the logic to work anyhow. It was a mistake I didn't do it from the beginning.

I don't see debugging value with having another goroutine wait and does nothing. Can you describe a specific case where it would be useful? You can pass the goproxy.Ctx to the CopyAndClose so that you'll know from goroutine dump which request context is it. Having channel makes the code slower and longer with no real benefit IMHO.

Does my two changes sound reasonable? I'll be happy to merge it then.

ap4y:
I changed my commit, but I still see some benefits of having a wait:

- It protects code against situation when socket will be improperly managed. Imagine that `handleHttps` returned and server crashes while goroutines are still processing. Ideally I think this type of errors should be gracefully handled but having wait will be a good start
- Code has a different routes in `handleHttp` (via `switch`) and only `ConnectAccept` shoulder returns without finishing operation. Not a biggie, but I think it might be nice to have more uniform behaviour here

I understand that waiting will eat some time, but having more determenistic logic might be worth it.

PR #162, opened 2016-05-18
Make it possible to disable outbound requests

boxofrad:
This PR introduces a `DisableOutboundRequests` option on `ProxyHttpServer` which prevents unmatched requests being forwarded on.

Example:

```Go
proxy.DisableOutboundRequests = true
proxy.OnRequest(goproxy.DstHostIs("www.reddit.com")).DoFunc(
    func(r *http.Request, _ *goproxy.ProxyCtx) (*http.Request, *http.Response) {
        return nil, goproxy.NewResponse(r, "text/plan", http.StatusOK, "Hello")
    }
)
```

Request 1:

    GET / HTTP/1.1
    Host: www.reddit.com

Response 1:

    HTTP/1.1 200 OK

    Hello

Request 2:

    GET / HTTP/1.1
    Host: www.google.com

Response 2:

    HTTP/1.1 404 Not Found

    goproxy: outbound requests have been disabled

This is useful for automated test suites where goproxy may be used to mock external APIs, but it is not desirable for HTTP requests to reach the real services.

Thanks for your stellar work supporting this project! 💖

### DIFF
--- a/https.go
+++ b/https.go
@@ -184,6 +184,11 @@ func (proxy *ProxyHttpServer) handleHttps(w http.ResponseWriter, r *http.Request
 
 				req, resp := proxy.filterRequest(req, ctx)
 				if resp == nil {
+					if proxy.DisableOutboundRequests {
+						sayOutboundRequestsDisabled(rawClientTls, ctx)
+						return
+					}
+
 					if err != nil {
 						ctx.Warnf("Illegal URL %s", "https://"+r.Host+req.URL.Path)
 						return
@@ -245,6 +250,12 @@ func (proxy *ProxyHttpServer) handleHttps(w http.ResponseWriter, r *http.Request
 			}
 		}
 		proxyClient.Close()
+	}
+}
+
+func sayOutboundRequestsDisabled(w io.WriteCloser, ctx *ProxyCtx) {
+	if _, err := io.WriteString(w, "HTTP/1.1 404 Not Found\r\n\r\n"+outboundRequestsDisabled); err != nil {
+		ctx.Warnf("Error responding to client: %s", err)
 	}
 }
 
--- a/proxy.go
+++ b/proxy.go
@@ -11,6 +11,10 @@ import (
 	"sync/atomic"
 )
 
+const (
+	outboundRequestsDisabled = "goproxy: outbound requests have been disabled\n"
+)
+
 // The basic proxy type. Implements http.Handler.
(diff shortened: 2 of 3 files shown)


elazarl:
Aren't we better off adding a filter that does just that?

Something like

proxy.OnRequest( /* whatever */ }
// but finally - block all requests that weren't handled.
proxy.OnRequest(func (req *http.Request) { return goproxy.NewResponse("no
dispatch") }

boxofrad:
Thanks @elazarl I do agree this would be cleaner...

However this relies on the blocking filter being added last, which in our (admittedly niche) use-case isn't straightforward.

For context: we have a project (soon to be open source) that allows you to control goproxy via an HTTP API, throughout our automated test suite we registers filters to mock responses from 3rd party APIs (by making HTTP requests to this API) so filters are added over time as apposed to up front.

Does that make sense? Do you have any other recommendations?
Thanks again!

elazarl:
*But in your project, you control the proxy, right?*

*So you can make sure the final filter would always be last, say, by
wrapping the proxy OnResponse methods with your own, which would always
make sure your response is last.*

*Maybe what you need is simply a OnResponeFirst() method that would make
sure you won't clobber the last filter?*

rusenask:
@boxofrad - in hoverfly (https://github.com/SpectoLabs/hoverfly) during 'simulate' mode after checking whether there are no matching request/responses  I just return an error that this was not found. In my opinion it's better to return an expected error so your automated test tools would identify it - https://github.com/SpectoLabs/hoverfly/blob/master/models.go#L575

boxofrad:
Thanks @elazarl I think you're right 😄 

What do you think would a good public API would be for this?
`OnResponseFirst` would get the job done, but feels like a strange method for the goproxy project as a whole outside the context of this PR.

Is there a way of achieving what I want to do without modifying goproxy at all?

boxofrad:
Also thanks for the suggestion @rusenask, that seems sensible!

Looking at your project it seems you're handling the request matching yourself (not using the condition arguments to `OnRequest` / `OnResponse`) is that right? 😃

rusenask:
hey, using `OnRequest` for 'hooking' into proxy :)

boxofrad:
I'm gonna close this PR as I've opted to using the same approach as hoverfly (using `OnRequest` for hooking into the proxy, and manually doing request matching).

Thanks for your help @rusenask and @elazarl ❤️

PR #166, opened 2016-05-24
prevent leaks by unblocking paired go routine

drewwells:
Closing the reader, can leave the pair go routine io.Copy hanging reading from the server net.Conn. By closing the writer, the io.Copy in the other routine will error out and unblock the goroutine. From debugging, proxy communications seems to happen this way:

1. Client: Hey Server
2. Server: I ack you
3. Client: send some data
4. Server: respond with data
5. Client: send some more data
6. Server: response with data
7. Client: close()
8. Server: hang

In other words, the server copy to client can hang forever under certain scenarios eventually causing the proxy to become unresponsive. I was able to reproduce this by chaining two proxies together, maybe it needs a test to simulate this tricky behavior?

### DIFF
--- a/https.go
+++ b/https.go
@@ -263,7 +263,7 @@ func copyAndClose(ctx *ProxyCtx, w, r net.Conn) {
 		connOk = false
 		ctx.Warnf("Error copying to client: %s", err)
 	}
-	if err := r.Close(); err != nil && connOk {
+	if err := w.Close(); err != nil && connOk {
 		ctx.Warnf("Error closing: %s", err)
 	}
 }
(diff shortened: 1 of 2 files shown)


elazarl:
Is it still needed after https://github.com/elazarl/goproxy/pull/161

drewwells:
I will have to test this, since it's similar to what I encountered as a fix. I don't think these low level calls are needed. From my experience, these go routines hang on reads not writes. Specifically reading server responses, since it never sends an `EOF`. I don't fully understand the use of the low level TCP closes, their use is discouraged https://golang.org/pkg/net/#TCPConn.CloseWrite

drewwells:
been using it a few hours, works well

PR #167, opened 2016-05-30
Close Body for MitM'ed connections

alxchk:
Replaced body of responses which are processed via proxy.OnRequest().HandleConnect(goproxy.AlwaysMitm) are never closed.


elazarl:
Doesn't `chunk.Close` close the response?

alxchk:
I don't know how these things works, but it doesn't. I have code which save hash sum of data on Body Close. For MitMed HTTPS it wasn't called. So I did this and it works now (at least for me) :)

PR #168, opened 2016-05-30
Fix panic if CONNECT transport is not a net.TCPConn

gwatts:
handleHttps would panic if Dial returned something other than a
net.TCPConn which was then used for a non MITM CONNECT request.

For example the golang.org/x/crypto/ssh package returns a net.Conn
when opening a tunnel through an ssh connection.  This is backed
by a wrapped ssh.Channel which also implements CloseWrite() so using
a type assertion achieves the same result in a portable fashion.

### DIFF
--- a/https.go
+++ b/https.go
@@ -99,10 +99,8 @@ func (proxy *ProxyHttpServer) handleHttps(w http.ResponseWriter, r *http.Request
 		ctx.Logf("Accepting CONNECT to %s", host)
 		proxyClient.Write([]byte("HTTP/1.0 200 OK\r\n\r\n"))
 
-		tTargetSiteCon := targetSiteCon.(*net.TCPConn)
-		tProxyClient := proxyClient.(*net.TCPConn)
-		go copyAndClose(ctx, tTargetSiteCon, tProxyClient)
-		go copyAndClose(ctx, tProxyClient, tTargetSiteCon)
+		go copyAndClose(ctx, targetSiteCon, proxyClient)
+		go copyAndClose(ctx, proxyClient, targetSiteCon)
 	case ConnectHijack:
 		ctx.Logf("Hijacking CONNECT to %s", host)
 		proxyClient.Write([]byte("HTTP/1.0 200 OK\r\n\r\n"))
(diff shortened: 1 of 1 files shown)


gwatts:
fwiw here's my program that uses goproxy to tunnel http requests over an SSH connection without requiring openssh, putty, etc https://github.com/gwatts/sshproxy

Other than the small patch, goproxy made it very easy to write the above, so thanks for that :-)

elazarl:
Hi,

My problem with it is, that I'm not sure if the logic would work correctly if you completely close the connection.

Wouldn't a better idea would be to accept a goproxy.CloseWriter interface, and then in your code you'll wrap net.Conn with a CloseWriter implementation?

```go
interface CloseReadWriter {
   CloseWrite() error
   CloseRead() error
}

type closeWrite struct {
    Closer io.Closer
}

func (c *closeReadWriter) CloseWrite() { c.Closer.Close() }
func (c *closeReadWriter) CloseRead() { c.Closer.Close() }

sshConn := CloseReadWriter{ssh.Dial(...)}
```

What do you think?

gwatts:
I agree that Close is probably not going to do the right thing within that call

OTOH even the http.Hijacker interface only returns a net.Conn; there's no guarantee that has a CloseWrite either; perhaps it should just test for CloseWrite and use that if available, else assume that writes are unbuffered and synchronous (net/http appears to follow this practice)

Additionally the code currently spawns those two goroutines to do the copy, but doesn't block on them, so  handleHttps is going to exit while the copy is still in progress, which doesn't seem ideal.

Finally it should still ensure Close is called on both connections before the handler exits, which doesn't appear to be the case currently.

I'm happy to update the PR with a rethought implementation if you agree with the above

elazarl:
How can you implement tunneling robustly without CloseWrite?

Regarding to the fact that the copy goroutine outlive the request handler. This is by design.
What benefit do you see for keeping the handler goroutine alive? The tunneling could be literally kept for hours.

elazarl:
@gwatts any update?

elazarl:
@gwatts can you please see if de25c6ed252fdc01e23dae49d6a86742bd790b12 fixes it for you?

PR #172, opened 2016-06-13
close response bodies for reuse in http package

drewwells:
HTTP responses should be closed when done using them. It can dramatically improve the performance of HTTP requests/responses.


drewwells:
This was the best post I could find on the topic: http://engineering.rainchasers.com/golang/2015/03/03/memory-leak-missing-body-close.html

elazarl:
@drewwells thanks. Great job!

PR #181, opened 2016-08-19
Fix `go vet` warnings

jabley:
These were noticed using Go 1.7. Earlier versions may not have had all
of these issues.
- https.go:376: assignment copies lock value to config:
  crypto/tls.Config contains sync.Once contains sync.Mutex
- examples/goproxy-yui-minify/yui.go:79: io.LimitedReader composite
  literal uses unkeyed fields
- proxy_test.go:508: unreachable code
- transport/transport.go:308: unreachable code
- transport/transport.go:658: unreachable code

### DIFF
--- a/https.go
+++ b/https.go
@@ -373,7 +373,7 @@ func (proxy *ProxyHttpServer) NewConnectDialToProxy(https_proxy string) func(net
 
 func TLSConfigFromCA(ca *tls.Certificate) func(host string, ctx *ProxyCtx) (*tls.Config, error) {
 	return func(host string, ctx *ProxyCtx) (*tls.Config, error) {
-		config := *defaultTLSConfig
+		config := defaultTLSConfig
 		ctx.Logf("signing for %s", stripPort(host))
 		cert, err := signHost(*ca, []string{stripPort(host)})
 		if err != nil {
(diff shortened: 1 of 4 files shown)


https.go
```diff
@@ -373,14 +373,14 @@ func (proxy *ProxyHttpServer) NewConnectDialToProxy(https_proxy string) func(net
 
 func TLSConfigFromCA(ca *tls.Certificate) func(host string, ctx *ProxyCtx) (*tls.Config, error) {
 	return func(host string, ctx *ProxyCtx) (*tls.Config, error) {
-		config := *defaultTLSConfig
```
elazarl: Isn't that a bug? You leak the defaultTLSConfig, and users can change it by mistake.

elazarl:
Hi!
Did you see my comment?
Am I correct?

jabley:
[I think using this should address it](https://go.googlesource.com/go/+/d24f446a90ea94b8), once Go 1.8 is out.

elazarl:
Can you please update the PR to clone the tls.Config?

Thanks

jabley:
Will do, once Go 1.8 is out in January 2017.

ErikPelli:
Fixed in https://github.com/elazarl/goproxy/pull/595

Issue #92, opened 2015-03-03
TestNoProxyHeaders and TestNoProxyHeadersHttps tests failing


glycerine:
I pulled the latest goproxy (99a857cc920470279e92a925f082a73fd4cf8dee), and the test suite exhibits two broken tests.

The broken tests are TestNoProxyHeaders and TestNoProxyHeadersHttps.

go1.4 darwin/amd64 on OSX 10.9.4

```
Jasons-MacBook-Pro:goproxy jasona$ go build
Jasons-MacBook-Pro:goproxy jasona$ go install
Jasons-MacBook-Pro:goproxy jasona$ go test -v
=== RUN TestSingerTls
--- PASS: TestSingerTls (0.17s)
=== RUN TestSingerX509
--- PASS: TestSingerX509 (0.15s)
=== RUN TestCounterEncDifferentConsecutive
--- PASS: TestCounterEncDifferentConsecutive (0.03s)
=== RUN TestCounterEncIdenticalStreams
--- PASS: TestCounterEncIdenticalStreams (0.00s)
=== RUN TestCounterEncStreamHistogram
--- PASS: TestCounterEncStreamHistogram (0.00s)
    counterecryptor_test.go:95: ref:20.09936102779116 - act:19.32792351627511 = 0.7714375115160514
=== RUN TestSimpleHttpReqWithProxy
2015/03/02 16:51:18 [003] WARN: Error copying to client: read tcp 127.0.0.1:52459: use of closed network connectio\
n
--- PASS: TestSimpleHttpReqWithProxy (0.01s)
=== RUN TestSimpleHook
--- PASS: TestSimpleHook (0.00s)
=== RUN TestAlwaysHook
--- PASS: TestAlwaysHook (0.00s)
=== RUN TestReplaceResponse
--- PASS: TestReplaceResponse (0.00s)
=== RUN TestReplaceReponseForUrl
--- PASS: TestReplaceReponseForUrl (0.00s)
=== RUN TestOneShotFileServer
--- PASS: TestOneShotFileServer (0.01s)
=== RUN TestContentType
--- PASS: TestContentType (0.00s)
=== RUN TestConstantImageHandler
--- PASS: TestConstantImageHandler (0.01s)
=== RUN TestImageHandler
--- PASS: TestImageHandler (0.01s)
=== RUN TestChangeResp
--- PASS: TestChangeResp (0.00s)
=== RUN TestReplaceImage
--- PASS: TestReplaceImage (0.01s)
[...]

PR #215, opened 2017-04-08
Fix TestNoProxyHeaders and TestNoProxyHeadersHttps tests

erikdubbelboer:
When the original request has a `Connection: close` header Go sets `r.Close` to true. When this is true Go adds the `Connection: close` header to the outgoing request even if the header was removed.

Fixes https://github.com/elazarl/goproxy/issues/92

This does mean that there is currently no way to make goproxy set a `Connection: close` header for its outgoing requests. I guess this could be made an option in `ProxyCtx` so you can change it in a request handler.


erikdubbelboer:
@elazarl no?

erikdubbelboer:
Tests are still failing...

ErikPelli:
Tests are not failing anymore, closing this PR

PR #218, opened 2017-04-13
Update the certificate files to match the contents of certs.go.

### LINKED ISSUE #216: CA certificate files ca.pem and key.pem no longer match cert.go
I assume that the ca.pem and key.pem files in the root are supposed to be those corresponding to what is in certs.go, but they no longer match because of https://github.com/elazarl/goproxy/commit/c4fc26588b6ef8af07a191fcb6476387bdd46711. Should ca.pem and key.pem be updated to match, or perhaps removed if we are supposed to copy/paste the cert info from certs.go when obtaining the cert to install in a client?

fredcy:
This fixes #216


elazarl:
Many thanks @fredcy !

PR #238, opened 2017-08-25
support inbounce & outbounce bandwidth and #OnDone

ddo:
Collect all the inbounce & outbounce bandwidth of http and https requests
This is real bandwidth, consist of the header and body not like the ``HandleBytes`` which only return the body of none ssl requests

Add ``#OnDone`` which will be trigger when a request done (http/https too)

So an example of bandwidth of all the request which go through our proxy:
```go
func main() {
	proxy.Verbose = true

	proxy.OnDone().DoFunc(func(ctx *goproxy.ProxyCtx) {
		fmt.Println("in :", ctx.Inbounce)
		fmt.Println("out:", ctx.Outbounce)
	})

	err := http.ListenAndServe(":9000", proxy)
	if err != nil {
		panic(err)
	}
}
```

### DIFF
--- a/actions.go
+++ b/actions.go
@@ -55,3 +55,17 @@ type FuncHttpsHandler func(host string, ctx *ProxyCtx) (*ConnectAction, string)
 func (f FuncHttpsHandler) HandleConnect(host string, ctx *ProxyCtx) (*ConnectAction, string) {
 	return f(host, ctx)
 }
+
+// OnDone is called when a http request complete including https
+type DoneHandler interface {
+	Handle(ctx *ProxyCtx)
+}
+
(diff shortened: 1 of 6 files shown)


elazarl:
Hi,
I'm not sure what are you trying to count.

Isn't that better to count on socket level? I.e, replace `net.Conn` with something that counts `Read` and `Write` and delegates to `net.Conn`?

elazarl:
Hi,
Is this still relevant?

Can you have a look at my comment?

ddo:
Thanks, I measure the bandwidth by the user requests. But I think it is not in this repo scope.

PR #242, opened 2017-09-20
change log lib to github.com/sirupsen/logrus

netxfly:


elazarl:
Sorry, but at the moment I want to keep the core library with zero dependencies.

Thanks  for the PR anyhow!

I can definitely see how  logus is more useful than the regular log.

PR #245, opened 2017-10-02
Update certificates used in test

ChimeraCoder:
The previous certificates were not actually valid, though older
versions of Go would accept them. See https://github.com/kubernetes/kubernetes/issues/38228#issuecomment-280213655

r? @carleton-stripe
cc @stripe/observability


elazarl:
This is already solved, right?

ChimeraCoder:
Yes - I meant to create the pull request against the `master` branch on our fork, not the upstream project. Sorry about that.

PR #248, opened 2017-10-10
Add tls.Certificate to proxy struct

astoliarov:
This pull request add tls.Certificate field to proxy struct. This prevent goproxy users from changing global ConnectAction's and GoproxyCA when they need to set custom certificate.
Also add helper: ParseCa create tls.Certificate from cert and key files


elazarl:
Can't you use the https function for that?

astoliarov:
Maybe. Could you give an example of code which do the thing?

astoliarov:
For me it was hard to understand where I need to put my own certificate when I start looking at goproxy and I found best explanation in customca example. This patch make it easier.
Also, after this patch we can run multiple proxies with different ca inside one app

astoliarov:
@elazarl sorry for pinging, but maybe you can review my PR again?

Thanks

astoliarov:
Found how to do the thing.

PR #256, opened 2017-11-08
Close the connection used by https 

YushiIso:
a socket of the connection created by https request remains.
Read and Write Connection will  be shutdown if the connection is rebuilt or  the connection is terminated.
however, the socket used by connection is not closed.
you can find it by using the "lsof" command.

the remaining connection is closed by time out, but I think that it is a problem if "high traffic" or "many clients use it".
it is better to close the used connection.


elazarl:
Can you write a test demonstrating that? So that it fails before your commit and succeed afterwards?

oec:
@elazarl, I can confirm this issue.  I created a test for this (on linux) that will fail before the commit and succeed afterwards: https://gist.github.com/oec/34d6acf1457b779ed7e446ddf3275762.

oec:
That would be @YushiIso's job, I suppose, as he owns the PR.

elazarl:
If you can send it as a difference PR, I'll merge both

YushiIso:
@oec  Thanks!!
Added test to the PR.

oec:
@elazarl, when is the best moment for the merge, if not now? :)

ErikPelli:
Can someone fix the issues with the tests? Is this problem still actual?

PR #264, opened 2018-01-15
Close connections that are not net.TCPConn properly

jasonzimmerman:
When using a Dialer that does not return net.Conn or net.TCPConn, connections are not closed properly when making an HTTP CONNECT request. This seems to stem from [https.go:113](https://github.com/elazarl/goproxy/blob/master/https.go#L113). It seems like the code is waiting for both connections to be closed before closing both connections. 
I fixed this in the pull request by changing the sync.WaitGroup to wait for one connection to close before closing the other one. Now the application will close the other end of the connection if either end closes their connection.

### DIFF
--- a/https.go
+++ b/https.go
@@ -110,7 +110,7 @@ func (proxy *ProxyHttpServer) handleHttps(w http.ResponseWriter, r *http.Request
 		} else {
 			go func() {
 				var wg sync.WaitGroup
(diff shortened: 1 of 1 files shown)


jasonzimmerman:
This causes a panic because the WaitGroup counter goes negative.

PR #266, opened 2018-01-19
original ctx.UserData is needing in new ctx variable when MITM

orkunkaraduman:
We need this to track the request between "HandleConnect()" and "Do()".


elazarl:
Wow, what a stupid bug in my part.

Sorry it took me so long.

PR #281, opened 2018-04-26
Keep Response Headers

gambol99:
The present behaviour is the drop all headers in the http.Response, which is our case causes issues with a set-cookie done before the proxy takes over. This PR adds an optional flag keeping the same behavior but allowing those to override if required


gambol99:
hi @elazarl ... any chance you can tell a look at this one :-)

gambol99:
Awesome ... thank you kindly @elazarl :-)

PR #284, opened 2018-05-13
Add possibility to install a cert store

btwotch:
Add the possibility to install a certificate storage to cache
certificates; this can make goproxy faster.

### DIFF
--- a/ctx.go
+++ b/ctx.go
@@ -1,6 +1,7 @@
 package goproxy
 
 import (
+	"crypto/tls"
 	"net/http"
 	"regexp"
 )
@@ -19,12 +20,18 @@ type ProxyCtx struct {
 	// call of RespHandler
 	UserData interface{}
 	// Will connect a request to a response
-	Session int64
-	proxy   *ProxyHttpServer
+	Session   int64
+	certStore CertStorage
+	proxy     *ProxyHttpServer
 }
 
 type RoundTripper interface {
 	RoundTrip(req *http.Request, ctx *ProxyCtx) (*http.Response, error)
+}
+
+type CertStorage interface {
+	Store(hostname string, cert *tls.Certificate)
+	Load(hostname string) *tls.Certificate
 }
 
 type RoundTripperFunc func(req *http.Request, ctx *ProxyCtx) (*http.Response, error)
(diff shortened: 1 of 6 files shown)


elazarl:
Can you explain the motivation for this PR? What does it enable you to do?

btwotch:
Hello @elazarl,
sorry for my late reply. With this change, I as a user of the goproxy framework, can use my own certificate store. I can decide for how long I want to cache them, whether I want to save them on hard drive or only keep them in memory. Perhaps I even want to use them after a restart of my program.

In my concrete example, I am caching the last 1000 of them. By doing so the I could decrease cpu usage.

oec:
@btwotch, I support this PR, in fact I use it in my own project. Very simple and versatile solution, thanks! I suppose @elazarl would be more convinced if the PR comes along with a test. :) Could you write a test for it?

btwotch:
@oec Thank you very much for your suggestion. Sorry that it take so long.
I added a test in proxy_test.go.

oec:
After using the CertStore for a while, we ran into a concurrency-issue:  Suppose we have multiple concurrent requests coming in, for the same target-host which has no certificate in the store yet.  Some of those requests will get nil from CertStore.Load().  Then a certificate is generated _for_ _each_ of those requests and stored via CertStore.Store(hostname, cert).  Therefore, each of those requests will use a different certificate.  This is not an uncommon scenario.

I think a better design would be to allow to specify your own `signHost` function for goproxy to use.  That function can implement load/generate/store operations atomically, including the use of a certificate-cache.

btwotch:
But you can make CertStore.Load()/Store() as atomic as signHost() as well, can't you?

I.e. to make signHost with a certificate-cache atomic you have to introduce a lock or a lock-free data structure. The same is true for CertStore.Load()/Store().

oec:
Indeed, you could extend your patch to do the locking inside signHost. But then again, with all the changes necessary to signHost, I chose to make _it_ pluggable directly.

oec:
Not yet, the code is vendored in into a project for a client, alongside with other cherry-picked patches for goproxy. I thought of putting all of that into a forked version of goproxy, but probably not anytime soon.

hmarr:
This can probably be closed now that #332 is merged

btwotch:
Yes.

PR #285, opened 2018-05-26
fix hijack compatibility related to RSA key length and cert common name; add ConnectDialToProxy handler for NewConnectDialToProxyWithHandler

liudanking:
1. fix hijack compatibility related to RSA key length and cert common name
2. add ConnectDialToProxy handler for NewConnectDialToProxyWithHandler

### DIFF
--- a/https.go
+++ b/https.go
@@ -276,6 +276,10 @@ func dialerFromEnv(proxy *ProxyHttpServer) func(network, addr string) (net.Conn,
 }
 
 func (proxy *ProxyHttpServer) NewConnectDialToProxy(https_proxy string) func(network, addr string) (net.Conn, error) {
+	return proxy.NewConnectDialToProxyWithHandler(https_proxy, nil)
+}
+
+func (proxy *ProxyHttpServer) NewConnectDialToProxyWithHandler(https_proxy string, connectReqHandler func(req *http.Request)) func(network, addr string) (net.Conn, error) {
 	u, err := url.Parse(https_proxy)
 	if err != nil {
 		return nil
@@ -290,6 +294,9 @@ func (proxy *ProxyHttpServer) NewConnectDialToProxy(https_proxy string) func(net
 				URL:    &url.URL{Opaque: addr},
 				Host:   addr,
 				Header: make(http.Header),
+			}
+			if connectReqHandler != nil {
+				connectReqHandler(connectReq)
 			}
 			c, err := proxy.dial(network, u.Host)
 			if err != nil {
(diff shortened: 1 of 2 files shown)


liudanking:
@elazarl 
Could you please have a review of this PR?
I have projects rely on this PR.
Thanks in advance.

signer.go
```diff
@@ -64,14 +64,15 @@ func signHost(ca tls.Certificate, hosts []string) (cert tls.Certificate, err err
 			template.IPAddresses = append(template.IPAddresses, ip)
 		} else {
 			template.DNSNames = append(template.DNSNames, h)
+			template.Subject.CommonName = h
```
elazarl: Why do you need to set the CommonName?
Isn't that deprecated in new certificates?
It doesn't seem like CreateCertificate uses CommonName.

signer.go
```diff
@@ -64,14 +64,15 @@ func signHost(ca tls.Certificate, hosts []string) (cert tls.Certificate, err err
...
 		}
 	}
 	var csprng CounterEncryptorRand
 	if csprng, err = NewCounterEncryptorRandFromKey(ca.PrivateKey, hash); err != nil {
 		return
 	}
 	var certpriv *rsa.PrivateKey
-	if certpriv, err = rsa.GenerateKey(&csprng, 1024); err != nil {
+	if certpriv, err = rsa.GenerateKey(&csprng, 2048); err != nil {
```
elazarl: It seems like a good idea, IIUC, chrome now rejects 1024 bits certs?

https.go
```diff
@@ -276,6 +276,10 @@ func dialerFromEnv(proxy *ProxyHttpServer) func(network, addr string) (net.Conn,
 }
 
 func (proxy *ProxyHttpServer) NewConnectDialToProxy(https_proxy string) func(network, addr string) (net.Conn, error) {
+	return proxy.NewConnectDialToProxyWithHandler(https_proxy, nil)
+}
+
+func (proxy *ProxyHttpServer) NewConnectDialToProxyWithHandler(https_proxy string, connectReqHandler func(req *http.Request)) func(network, addr string) (net.Conn, error) {
```
elazarl: Can you add doc/example how should this be used?

signer.go
liudanking: Some softwares (e.g., okhttp) still check CommonName when verifying cert.
This is for forward compatibility.

signer.go
liudanking: More and more libs and softwares reject cert less than 1024 bits now, 2048 is a better option.

https.go
liudanking: Example added, please have a look.

liudanking: @elazarl 😄

PR #291, opened 2018-06-16
Authentication information might be used by request/connect handlers later

onokonem:
Hi,

For our own needs we've adjusted goproxy to store auth info from basic auth handler to ProxyCtx to provide it to other handlers later.

Please accept if it's ok

Please let me know if it has to be modified somehow to be accepted.

Thank you.

With best regards,
Daniel Podolsky


ErikPelli:
This pull request doesn't add anything that can't be already done without these changes.
`User` and `Passowrd` can be persisted inside the `UserData` custom field.
- ConnectDial and Logger are already present in master branch
- OnDone() handler is useless and just a Response Handler can do the same, since it's sent when the request is finished
- The custom resolver can be already be used with a custom `net.Dialer` (e.g. https://koraygocmen.medium.com/custom-dns-resolver-for-the-default-http-client-in-go-a1420db38a5d)

I'm closing this Pull Request.
If you have any other proposal, open another PR, please!

PR #298, opened 2018-07-11
Fix issue 297: TestCounterEncStreamHistogram broken in go1.11.

ijt:
This doubles the number of samples to reduce the variability of the stddev estimates.


elazarl:
Thanks!

PR #302, opened 2018-07-27
add set logger

zhizhiboom:
Modifying logger supports third-party logging packages


elazarl:
Hi,

What do you think about my solution of using a custom log function?

And then, say, add to `ext` a function that uses logrus, or whatever logger you choose?

elazarl:
Solved by recent changes of logger to be an interface.

Issue #311, opened 2018-09-18
Firefox reported SEC_ERROR_REUSED_ISSUER_AND_SERIAL for MITM


ShichangXu:
I run examples/goproxy-eavesdropper/main.go to perform MITM on HTTPS connections.
I installed the certificate in firefox and open web pages through goproxy.

The webpages are not rendered properly. I use the developer tools in firefox to examine the network activities and found that ~50% of objects are not succeeded downloaded. Firefox reported SEC_ERROR_REUSED_ISSUER_AND_SERIAL for those failed objects. This seems to be issue with the certificates.

I use wireshark to capture the generated certificates from goproxy. It seems that goproxy can generates different certificates even for the same domain. This causes issue for firefox. I guess that firefox will check and make sure the certificate for each domain is constant.

I checked the goproxy code. In signer.signHost(), it calls x509.CreateCertificate(), which eventually calls randutil.MaybeReadByte(). This function generates some randomness and reads a single byte from r with ~50% probability. It seems that this is causing goproxy to generate different certificates for the same domain.

I implemented a quick fix of the code and let goproxy to cache the generated certificate for each domain. This seems to fix this issue.

I am not familiar with Go language that much and not very confident what I found is correct. Could you help me confirm and fix this? Thanks.

anhhh11:
Cached in the go dictionary not enough; when you restart your goproxy application it sometime will appear SEC_ERROR_REUSED_ISSUER_AND_SERIAL error again. I've implemented cached to when open app read to the dictionary for the first time from file; the next request will read from the dictionary; this solution seem very stable not seeing SEC_ERROR_REUSED_ISSUER_AND_SERIAL again.

ShichangXu:
Thanks for the quick response. This solution makes sense.

However, after I changed the code, I cleared the cache and restarted firefox, some webpages are still not rendered properly. For example, all images on Amazon are not shown, as the domain "images-na.ssl-images-amazon.com" still reports SEC_ERROR_REUSED_ISSUER_AND_SERIAL.

I am wondering what is the fundamental cause of this issue. Are there other reasons why this issue is triggered.

anhhh11:
It seems like SSLs are cached; try to clear history:
https://stackoverflow.com/questions/24044392/how-do-you-clear-an-ssl-state-in-a-firefox-version-20-web-browser

ShichangXu:
Thanks! This might be the cause.

PR #314, opened 2018-10-10
random serial

eternal-flame-AD:
I am still experiencing the same issue with #311. The problem can be reproduced by:
1. start the proxy server
2. open firefox and view a hijacked page
3. restart the proxy server(which means all cached certificates at the server-side are cleared)
4. view the same page without closing the browser

I think the problem is cause by a misuse of hash for certificate serial number. As each time `signHost` is called, a different certificate is generated **but the serial is gonna be the same**, which would trigger an alert in Firefox.
I think it would be better to simply generate a random serial number.

### DIFF
--- a/signer.go
+++ b/signer.go
@@ -1,6 +1,7 @@
 package goproxy
 
 import (
+	"crypto/rand"
 	"crypto/rsa"
 	"crypto/sha1"
 	"crypto/tls"
@@ -8,7 +9,6 @@ import (
 	"crypto/x509/pkix"
 	"math/big"
 	"net"
-	"runtime"
 	"sort"
 	"time"
 )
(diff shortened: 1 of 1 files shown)


elazarl:
How would you keep it consistent between runs?

eternal-flame-AD:
We do not need to keep it consistent between runs. What we needed to prevent is generating certificates with the same serial number, even between runs. I cannot see a side-effect as long as the serial numbers are kept different from each other.

iwex:
@elazarl Hello! Also interested in some solution... Thanks!

eternal-flame-AD:
Cleaning my own stale PRs, looks like solved in #355

Issue #102, opened 2015-05-18
Removing Accept-Encoding header


timrwilliams:
I'm running goproxy as a simple HTTP & HTTPS proxy, just relaying requests and responses as is without modifying or inspecting the content.

I've noticed that you're stripping out the Accept-Encoding header on all requests presumably to allow us to inspect the response without having to decode it. Is it feasible to leave this header intact if we only want to relay the response back to the client?

Essentially I want to allow a client to specify a header of "Accept-Encoding: gzip, deflate" and pass this through to the end server.

elazarl:
I do not do that, it is the Go http server that does, since it want to
decode the gzip'd stream itself.

If you want to be a fully transparent proxy, why do you need goproxy at all?

I'm not saying this is not a good idea, I'm just trying to understand the
use case.

timrwilliams:
I'm collecting stats on traffic as it passes through the proxy so we inspect the headers on the way in and out but not the body content itself.

I saw the header deletion here and assumed that you'd made a design decision to remove it:
https://github.com/elazarl/goproxy/blob/master/proxy.go#L79

timrwilliams:
I've looked in to it a bit further and understand what you mean now, I didn't realise transport.go was just the standard golang library.

I've commented out the header deletion line in proxy.go in my local copy of your repo and its working fine for my use case. All your test cases pass too.

Looking at a few other proxy implementations and it seems a common approach is to offer a boolean configuration option that will remove or preserve this header. 

I'm happy using my modified fork if you don't think my use case justifies a change to core code.

elazarl:
I'm really not sure what to do.

The problem is, not removing these header might cause problems. For example
client might accept an uncommon encoding, not supported by Go. When the
user would send it, he wouldn't be able to read the body.

Maybe we can keep the original headrs, and add a hook that would allow you
to add them back before sending them to the server.

What do you think?

timrwilliams:
I don't think removing and re-adding the Accept-Encoding header would work as it changes how a client and server encode the response so can't be added on at the end.

To be honest I don't know enough of the inner workings of the Go transport layer to know what edge cases could cause my approach to go wrong.

I send a decent amount of real world traffic from different applications to different servers through my proxy so what I could do is put this change out there on one of my instances and see how it behaves in the wild. We can mark this issue as pending or closed and come back to it in the future if its an issue for other people or if I see any adverse results.

I really appreciate you looking in to this by the way.

rakoo:
[...]

PR #320, opened 2019-01-10
Add KeepAcceptEncoding option, prevents proxy from changing Accept-Encoding sent by clients

bls:
Hi,

This PR adds a new flag to the proxy, `KeepAcceptEncoding`. When disabled, proxy behaviour remains the same to avoid breaking existing clients.

The default behaviour is that the Accept-Encoding header on client requests is removed. The proxy outbound transport will then add an "Accept-Encoding: gzip" header to outbound requests.

With `KeepAcceptEncoding` enabled the proxy will pass the header through. Note that if you want the proxy to *never* modify Accept-Encoding, you should also set `DisableCompression` true on the proxy outbound transport (otherwise it will ask for gzip when the client does not send the header).

Fixes https://github.com/elazarl/goproxy/issues/102 for me.

### DIFF
--- a/proxy.go
+++ b/proxy.go
@@ -80,7 +80,7 @@ func removeProxyHeaders(ctx *ProxyCtx, r *http.Request) {
 	ctx.Logf("Sending request %v %v", r.Method, r.URL.String())
 	// If no Accept-Encoding header exists, Transport will add the headers it can accept
 	// and would wrap the response body with the relevant reader.
-	r.Header.Del("Accept-Encoding")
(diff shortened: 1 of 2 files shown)


elazarl:
I now wonder why I added accept encoding, and not just send client's accept encoding.

In retrospect it looks like a mistake.

What does squid do?

ErikPelli:
This feature has been integrated in https://github.com/elazarl/goproxy/pull/585. Closing this PR.

PR #323, opened 2019-01-20
Added logger interface to allow usage of custom loggers

mbilski:
This change allows usage of custom logging libraries like logrus etc instead of go standard log library.
In my case needed for logging warnings in json format.


moooofly:
need this too.

elazarl:
Looks good. Thanks!

PR #329, opened 2019-03-01
Prevent indirect passing of the Connection header by the proxy

tommysitu:
If `request.Close` is not set to false, the `transfer.writeHeader` function will still write the "Connection: close" header.


ErikPelli:
This change is currently present in the master branch, closing this PR

PR #331, opened 2019-03-31
Add go.mod to make project Go modules compatible

hmarr:
That.


elazar-lb:
Thanks! 
I really really wanted this kind of PR!

One question.
Can we push these dependencies to the examples dir? So that goproxy itself would remain dependency less?

hmarr:
Good point - I hadn't realised those dependencies were specific to the examples. Done!

elazarl:
@hmarr sorry to nag, but shouldn't the `go.sum` at root be empty as well?

hmarr:
Good point - fixed.

Running `go mod tidy` brought back the `github.com/rogpeppe/go-charset` dependency in the top-level `go.mod`, as the `ext/html` package depends on it, so I had to create an additional `go.mod` in `ext`.

PR #338, opened 2019-04-25
Speed up certificate generation time by reusing publicKey and priateKey.

zckevin:
Reduce function `signHost` execution time from N*100 ms to N ms.


elazarl:
Currently there are conflicts

PR #339, opened 2019-05-01
 Do not modify headers in response to a HEAD request

sburton84:
Currently GoProxy always removes the `Content-Length` header from any response, and replaces the body with chunked encoding. This does not really make sense to apply to `HEAD` requests, which will have no body in the response, and may rely on the `Content-Length` being returned.

Use-case: The Docker Registry API sends a `HEAD` request to determine if a blob for a particular layer already exists in the repository. It expects that, if the blob exists, the response will be a `200 OK` and will contain a `Content-Length` header indicating the size of the blob. If GoProxy strips the header, then Docker thinks that the blob does not exist, and always tries to reupload every blob. This change allows GoProxy to correctly proxy the Docker Registry API.

### DIFF
--- a/https.go
+++ b/https.go
@@ -234,32 +234,44 @@ func (proxy *ProxyHttpServer) handleHttps(w http.ResponseWriter, r *http.Request
 					ctx.Warnf("Cannot write TLS response HTTP status from mitm'd client: %v", err)
 					return
 				}
-				// Since we don't know the length of resp, return chunked encoded response
-				// TODO: use a more reasonable scheme
-				resp.Header.Del("Content-Length")
-				resp.Header.Set("Transfer-Encoding", "chunked")
-				// Force connection close otherwise chrome will keep CONNECT tunnel open forever
-				resp.Header.Set("Connection", "close")
-				if err := resp.Header.Write(rawClientTls); err != nil {
-					ctx.Warnf("Cannot write TLS response header from mitm'd client: %v", err)
-					return
-				}
-				if _, err = io.WriteString(rawClientTls, "\r\n"); err != nil {
-					ctx.Warnf("Cannot write TLS response header end from mitm'd client: %v", err)
-					return
-				}
-				chunked := newChunkedWriter(rawClientTls)
-				if _, err := io.Copy(chunked, resp.Body); err != nil {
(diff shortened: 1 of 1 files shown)


sburton84:
Looking further into this, I'm not sure I understand the need to convert to chunked encoding at all? What was the reasoning behind this? I don't see why we can't just write the response back to the client unchanged like so:

```
if err := resp.Write(rawClientTls); err != nil {
  ctx.Warnf("Cannot write TLS response from mitm'd client: %v", err)
  return
}		
```

I don't think it matters that we don't know the size of the body, it is an `io.Reader` so needn't be all read into memory in one go, and Go will parse it correctly whether it has `Content-Length` or is chunked and will return `EOF` appropriately when the end is reached.

elazarl:
@sburton84 the problem is you don't know in advance how much will be written

sburton84:
> @sburton84 the problem is you don't know in advance how much will be written

But why does it matter? The handler is running inside a goroutine and the reader and writer are streams so there's no reason you can't just stream from the reader to the writer until you reach EOF.

elazarl:
The only way to stream, is not to set the `Content-Length` header.

sburton84:
That's not true, streaming the data this way makes no difference to the client as it's just a stream of bytes, no different to how it would receive any HTTP body from the network whether it has a `Content-Length` header or not. The only difference the header makes is that without one you need to indicate the end of the body by closing the socket.

Either way, as long as you assume the upstream server is behaving correctly, there's no reason you can't forward whatever headers it provides then stream the body that it provides. Doing this the client will receive the exact same data that any direct client to the upstream server would have received.

RoganDawes:
HTTP is a bit annoying in its handling of the HEAD method.
Essentially you need to keep track of whether the request method was GET or HEAD. 
If the method is a GET, and the response has a Content-Length header, you should wrap the response content stream with an io.LimitedReader set to the value in the Content-Length header. 
However, if the request method was HEAD, and the response has a Content-Length header, you should wrap the response content stream with an io.LimitedReader set to 0. Or some other way to indicate that there is no actual content to be sent.

ErikPelli:
Fixed by https://github.com/elazarl/goproxy/pull/473

PR #340, opened 2019-05-07
Add support for signing with ecdsa keys

johnerikhalse:
Signing with Eliptic Curve DSA sertificates is a lot faster than using RSA.
This PR adds support for ECDSA and also adds benchmarks to show the performance win.

```
$ go test -bench=. -run=XXX -benchtime=30s
goos: linux
goarch: amd64
pkg: github.com/elazarl/goproxy
BenchmarkRsaTls-4            200         183597836 ns/op
BenchmarkRsaX509-4           300         178512072 ns/op
BenchmarkEcdsaTls-4        30000           1701068 ns/op
BenchmarkEcdsaX509-4      200000            340498 ns/op
PASS
ok      github.com/elazarl/goproxy      266.347s
```
By also adding PR #338, the benchmark numbers are as follows:

```
$ go test -bench=. -run=XXX -benchtime=30s
goos: linux
goarch: amd64
pkg: github.com/elazarl/goproxy
BenchmarkRsaTls-4           2000          19618000 ns/op
BenchmarkRsaX509-4          5000           9091853 ns/op
BenchmarkEcdsaTls-4        30000           1672054 ns/op
BenchmarkEcdsaX509-4      200000            325934 ns/op
PASS
ok      github.com/elazarl/goproxy      223.161s
```


elazar-lb:
As discussed, this is missing the point of having stable certificates for each site, given a stable CA.

johnerikhalse:
I have added back the predictable certificate generation. I also changed the benchmarks to only benchmark the signing method.

elazarl:
Thanks. I need to rethink the stable CA. Not sure if its worth it. It seems like the serial number is wide enough to allow GUID generation without fearing collision. Maybe that's what we should do.

PR #346, opened 2019-06-12
fixes #345: panic when resp == nil

### LINKED ISSUE #345: panics when cannot receive response
To reproduce:
1. Run the simple proxy example.
2. Use `wget` to send it a request to a website that is down.

Don't see a reason to panic, why just don't call the DoHandler if the proxy failed to get a response?

kubistika:
#345

### DIFF
--- a/proxy.go
+++ b/proxy.go
@@ -70,6 +70,10 @@ func (proxy *ProxyHttpServer) filterRequest(r *http.Request, ctx *ProxyCtx) (req
 func (proxy *ProxyHttpServer) filterResponse(respOrig *http.Response, ctx *ProxyCtx) (resp *http.Response) {
 	resp = respOrig
 	for _, h := range proxy.respHandlers {
+		if resp == nil {
+			// resp is nil, no need to run the next handler
+			return
+		}
 		ctx.Resp = resp
 		resp = h.Handle(resp, ctx)
 	}


elazar-lb:
Thanks,
Wouldn't it be better to handle `nil` yourself in the handler?

azak-azkaran:
> Thanks,
> Wouldn't it be better to handle `nil` yourself in the handler?

I think so too, but I think there is another bug corresponding to this one:

If the response is `nil` the following lines will end in a panic in Line [132](https://github.com/elazarl/goproxy/blob/6125c262ffb039dd0fdb9ba51987c0ac57b4a124/proxy.go#L132):
https://github.com/elazarl/goproxy/blob/6125c262ffb039dd0fdb9ba51987c0ac57b4a124/proxy.go#L130-L132

kubistika:
@elazar-lb @azak-azkaran that's the exact problem. Nil can not be handled by client, that why this pr is necessary

elazarl:
I see, is there a PR solving the panic only?

azak-azkaran:
I think if we move the error handling:
https://github.com/elazarl/goproxy/blob/6125c262ffb039dd0fdb9ba51987c0ac57b4a124/proxy.go#L121-L125

The solution becomes cleaner, what do you think? 
I added my proposed solution and a test case in the provided file: [345.txt](https://github.com/elazarl/goproxy/files/3358889/345.txt)

kubistika:
@azak-azkaran looks better, i'll update
And yes, it solves the panic

PR #348, opened 2019-07-05
Issue: #345 - fixes resp == nil:

azak-azkaran:
* fixes the panic if the response is from a handler is nil
* adds test case for handler response is nil


azak-azkaran:
I created a PR after the discussion in #346
I hope that is ok :)

PR #351, opened 2019-07-11
Adds cache for MITM signing to speed up repeat proxy requests.

kfreezen:
The goal of this PR was to reduce the number of times the proxy generated a leaf certificate for a website.

Feel free to pick it apart. I am using it in our custom version, so I thought I would create a PR to bring it upstream if it's useful.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,1 +1,3 @@
 module github.com/elazarl/goproxy
+
+go 1.12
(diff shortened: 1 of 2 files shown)


elazarl:
Can you solve the conflict in the `go.mod` file?

hazcod:
@kfreezen would be nice to make it an expiring cache to not keep certificates forever.

ErikPelli:
There is already CertStore that caches the generated certificates and reuse them for the same host in MITM.
You can use your own cache implementation with it.
Closing this pull request.

PR #355, opened 2019-07-24
Random cert serial

### LINKED ISSUE #311: Firefox reported SEC_ERROR_REUSED_ISSUER_AND_SERIAL for MITM

dmjones:
Creates a random certificate serial for each generated certificate. This prevents browsers seeing the same serial, for the same host, in two different certificates.

It is sufficient to use non-cryptographic random numbers for this task, hence the use of `math/rand`. If deterministic behaviour is needed (e.g. in tests), `rand.Seed()` can be called again.

This PR also includes some unavoidable changes to go.mod/go.sum files, without which the project cannot build (at least with Go 1.12).

Fixes #311.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,1 +1,3 @@
 module github.com/elazarl/goproxy
+
+require github.com/elazarl/goproxy/ext v0.0.0-20190711103511-473e67f1d7d2
--- a/signer.go
+++ b/signer.go
@@ -11,6 +11,7 @@ import (
 	"crypto/x509/pkix"
 	"fmt"
 	"math/big"
+	"math/rand"
 	"net"
 	"runtime"
 	"sort"
@@ -48,9 +49,8 @@ func signHost(ca tls.Certificate, hosts []string) (cert *tls.Certificate, err er
 	if err != nil {
 		panic(err)
 	}
-	hash := hashSorted(append(hosts, goproxySignerVersion, ":"+runtime.Version()))
-	serial := new(big.Int)
-	serial.SetBytes(hash)
+
+	serial := big.NewInt(rand.Int63())
 	template := x509.Certificate{
 		// TODO(elazar): instead of this ugly hack, just encode the certificate and hash the binary form.
 		SerialNumber: serial,
@@ -74,6 +74,7 @@ func signHost(ca tls.Certificate, hosts []string) (cert *tls.Certificate, err er
 		}
 	}
 
+	hash := hashSorted(append(hosts, goproxySignerVersion, ":"+runtime.Version()))
 	var csprng CounterEncryptorRand
 	if csprng, err = NewCounterEncryptorRandFromKey(ca.PrivateKey, hash); err != nil {
 		return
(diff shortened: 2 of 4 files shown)


go.mod
```diff
@@ -1 +1,3 @@
 module github.com/elazarl/goproxy
+
+require github.com/elazarl/goproxy/ext v0.0.0-20190711103511-473e67f1d7d2
```
elazarl: Why do we need this? main goproxy module shouldn't have dependency, should it?

signer.go
```diff
@@ -102,3 +103,8 @@ func signHost(ca tls.Certificate, hosts []string) (cert *tls.Certificate, err er
 		PrivateKey:  certpriv,
 	}, nil
 }
+
+func init() {
+	// Avoid deterministic random numbers
+	rand.Seed(time.Now().UnixNano())
```
elazarl: Is that the canonical Go way to seed rand?

elazarl:
Looks good overall. Can you have a look at the comments?

go.mod
dmjones: `github.com/elazarl/goproxy/ext` has its own `go.mod` file, which means its a standalone Go module. Since this module is consumed by your `github.com/elazarl/goproxy` package, a dependency is formed. I only performed this change because the current `go.mod` was invalid.

If you have no plans to use `github.com/elazarl/goproxy/ext` as a separate package, you could remove the `go.mod` file. But that's outside the scope of this PR.

signer.go
dmjones: I'm not sure if there is a canonical way. We do not need cryptographically strong random numbers (otherwise we'd use `crypto/rand`), sowe just need to make it unlikely that two consecutive runs of the code will produce the same string of random numbers.

If you look in the second example on https://golang.org/pkg/math/rand/, you'll see the following comment:

```go
	// Typically a non-fixed seed should be used, such as time.Now().UnixNano().
```

go.mod
elazarl: I understand. goproxy shouldn't depend on `goproxy/ext`. But as you said, this is a mistake for another PR.
Thanks.

Issue #50, opened 2014-05-06
Rejecting a CONNECT request with an html message?


ismail:
I am trying to Reject CONNECT requests via

```
proxy.OnRequest().HandleConnectFunc(func(host string, ctx *goproxy.ProxyCtx) (*goproxy.ConnectAction, string) {
            ctx.Resp = goproxy.NewResponse(ctx.Req,
                    goproxy.ContentTypeHtml, http.StatusForbidden,
                    "<br><br><center><h1>Address is blocked.</h1></center>")
            return goproxy.RejectConnect, host
}
```

But this doesn't seem to work, browser shows the connection problem page instead of the html.

osiloke:
+1

abourget:
@ismail took a look at `abourget/goproxy` ? you can short-circuit any handlers with a ctx.NewResponse() .. including the CONNECT handlers..

abourget:
@osiloke read previous comment too

PR #368, opened 2019-11-29
MITM: recreating private/public keys all the time

stiray:
... has huge impact on speed of processing, with this change the RSA and ECDSA keys are created within "init()" and used when needed. This is quick fix, please set it to some configuration variable if you consider that it is useful to generate them on each request. I dont think this would impact the security on client side.

### DIFF
--- a/signer.go
+++ b/signer.go
@@ -13,10 +13,13 @@ import (
 	"fmt"
 	"math/big"
 	mathrand "math/rand"
+	"runtime"
 	"net"
 	"sort"
+	"strings"
 	"time"
 	"github.com/patrickmn/go-cache"
+	"sync/atomic"
 )
 
 func hashSorted(lst []string) []byte {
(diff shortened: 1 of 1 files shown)


signer.go
```diff
@@ -36,8 +40,86 @@ func hashSortedBigInt(lst []string) *big.Int {
...
+	key := strings.Join(hosts[:], "/")
+	value, exist := crtCache.Get(key)
+	if exist {
+		cert = value.(*tls.Certificate)
+		return
+	}
+
+    cert, err = generateCertificate(ca, hosts)
+
+	crtCache.Add(key, cert, cache.DefaultExpiration)
```
engineering-this: Race condition scenario:

Two concurrent calls to signHost for hostname with not yet cached crt.
The second call will fail to add cache, as it is existing, although it didn't exist when line #50 was called.

hazcod: @engineering-this: how would you solve that race condition?

engineering-this: I'd go with a semaphore.

```go
var semaphore = make(chan struct{}, 1)

func signHost(ca tls.Certificate, hosts []string) (cert *tls.Certificate, err error) {
	semaphore <- struct{}{}
	defer func() {
		<-semaphore
	}()
	//...
```

hazcod:
@stiray @elazarl @engineering-this : see https://github.com/hazcod/goproxy/commit/152f8657271878d3f574290108e33e2d6b70c348 for a possibly better implementation.

stiray:
Ok, I am not using github and I dont even intend to as I have my own git server and I am quite happy with command line so I wont be clicking anything here any more... 

Regarding the "race condition" which was probably merged due to clicking to "resolve whatever" which was a mistake if the fix came into code.

The race condition was put there deliberately to avoid synchronizing multiple threads using synchronization object at a cost that here and then the certificate is generated without any reason which is performance wise better decision.

But nvm me, this was a drive by code, it have my own repo/fork so please do whatever feels appropriate for you.

ErikPelli:
Proxy server now has a `CertStore` field that caches the certificates and fix the issue that you explained. This merge request is not needed anymore.

PR #369, opened 2020-02-19
improvement: more general support for half-closable connections

onokonem:


https.go
```diff
@@ -64,6 +64,14 @@ func (proxy *ProxyHttpServer) connectDial(network, addr string) (c net.Conn, err
 	return proxy.ConnectDial(network, addr)
 }
 
+type halfClosable interface {
+	net.Conn
+	CloseWrite() error
+	CloseRead() error
+}
+
+var _ halfClosable = (*net.TCPConn)(nil)
```
elazarl: Why is that needed?

onokonem: this one is just to check the interface is complaint to the needs.

onokonem:
> Looks nice, but I wonder, what is the usecase?

actually, 2 cases (and I run into both in our project)

1. Not only net.TCPConn is half-closable, but also the net.UnixConn
2. sometimes we have to wrap net.TCPConn with some custom type (we are reporting traffic, for example)

PR #370, opened 2020-03-06
Make project pass `go vet`.

bartle-stripe:


bartle-stripe:
Thanks, removed the redundant comment.

PR #374, opened 2020-03-27
Fixed compile errors

mach-spotqa:


counterecryptor_test.go
```diff
@@ -9,7 +9,7 @@ import (
 	"math/rand"
 	"testing"
 
-	"github.com/elazarl/goproxy"
+	"github.com/mach-spotqa/goproxy"
```
elazarl, line 12: We should keep this as it is for this repo, right?

mach-spotqa: My bad. Included it on PR by mistake. But it should fix compile error as new PR #378

Issue #376, opened 2020-03-28
HandleHttps should not send 200 before calling ConnectHijack's Hijack func


riraccuia:
*ProxyHttpServer.handleHttps() currently writes a HTTP/1.0 200 OK back to the client connection before ConnectHijack's Hijack function is invoked. 

[RFC 2817](https://tools.ietf.org/html/rfc2817#section-5.3) States: 

```
5.3 Establishing a Tunnel with CONNECT

   Any successful (2xx) response to a CONNECT request indicates that the
   proxy has established a connection to the requested host and port,
   and has switched to tunneling the current connection to that server
   connection.
```

It should be therefore up to the Hijack() method implementer to send the http 200 upon confirming that the destination (wether it was changed or not) is reachable. 

This makes it impossible to cascade the http 502 back to the client in case we wanted to abort prematurely (e.g.the target address is unavailable) #376

riraccuia:
I created a pull request

riraccuia:
see #377

ErikPelli:
https://github.com/elazarl/goproxy/pull/377 fixed this

PR #377, opened 2020-03-28
Don't return http 200 for ConnectHijack actions

riraccuia:
#376  
Sending HTTP 200 to the client connection before invoking the todo.Hijack() method for a ConnectHijack action makes it impossible to cascade the http 502 (or any other error) back to the client in case we wanted to abort prematurely (e.g. the target address is unavailable)

[RFC 2817](https://tools.ietf.org/html/rfc2817#section-5.3) States: 

```
5.3 Establishing a Tunnel with CONNECT

   Any successful (2xx) response to a CONNECT request indicates that the
   proxy has established a connection to the requested host and port,
   and has switched to tunneling the current connection to that server
   connection.
```

It should be therefore up to the Hijack() method implementer to send the http 200 upon confirming that we can establish a tunnel.

### DIFF
--- a/https.go
+++ b/https.go
@@ -36,6 +36,10 @@ var (
 	httpsRegexp     = regexp.MustCompile(`^https:\/\/`)
 )
 
+// ConnectAction enables the caller to override the standard connect flow.
+// When Action is ConnectHijack, it is up to the implementer to send the 
+// HTTP 200, or any other valid http response back to the client from within the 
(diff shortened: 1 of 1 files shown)


dmagyar:
I would like to second this PR, we had to maintain a local fork so we can hijack 200 messages on CONNECTs. Thanks for raising this @riraccuia!

PR #380, opened 2020-04-24
Add StatusCodeIs response condition

n-i-x:
Adding a response condition for HTTP status codes.

Usage:

`proxy.OnResponse(statusCodeIs(301, 302, 303, 307, 308)).DoFunc(...`

### DIFF
--- a/dispatcher.go
+++ b/dispatcher.go
@@ -161,6 +161,22 @@ func ContentTypeIs(typ string, types ...string) RespCondition {
 	})
 }
 
+// StatusCodeIs returns a RespCondition, testing whether or not the HTTP status
+// code is one of the given ints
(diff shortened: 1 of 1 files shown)


hoozecn:
@elazarl Can't compile goproxy with this patch

dispatcher.go
```diff
@@ -161,6 +161,22 @@ func ContentTypeIs(typ string, types ...string) RespCondition {
 	})
 }
 
+// StatusCodeIs returns a RespCondition, testing whether or not the HTTP status
+// code is one of the given ints
+func StatusCodeIs(codes ...int) goproxy.RespCondition {
```
hoozecn, line 166: @n-i-x @elazarl `goproxy.` should be removed in this function

Issue #306, opened 2018-08-10
HTTP in cascadeproxy example returns Proxy Authentication Required


gbraad:
When setting

https://github.com/elazarl/goproxy/blob/947c36da3153ff334e74d9d980de341d25f358ba/examples/cascadeproxy/main.go#L85

as

```diff
-	request, err := http.NewRequest("GET", "https://ip.cn", nil)
+	request, err := http.NewRequest("GET", "http://ip.cn", nil)
```

execution results in

```
$ ./main
2018/08/10 05:37:35 serving end proxy server at localhost:8082
2018/08/10 05:37:35 serving middle proxy server at localhost:8081
2018/08/10 05:37:36 [001] INFO: Got request / ip.cn GET http://ip.cn/
2018/08/10 05:37:36 [001] INFO: Sending request GET http://ip.cn/
2018/08/10 05:37:36 [001] INFO: Got request / ip.cn GET http://ip.cn/
2018/08/10 05:37:36 [001] INFO: Copying response to client  [407]
2018/08/10 05:37:36 [001] INFO: Copied 33 bytes to client error=<nil>
2018/08/10 05:37:36 [001] INFO: Received response 407 Proxy Authentication Required
2018/08/10 05:37:36 [001] INFO: Copying response to client 407 Proxy Authentication Required [407]
2018/08/10 05:37:36 [001] INFO: Copied 33 bytes to client error=<nil>
2018/08/10 05:37:36 status 407, data 407 Proxy Authentication Required
```

While the headers are set in `middleproxy` using:

https://github.com/elazarl/goproxy/blob/947c36da3153ff334e74d9d980de341d25f358ba/examples/cascadeproxy/main.go#L74-L77

https://github.com/elazarl/goproxy/blob/947c36da3153ff334e74d9d980de341d25f358ba/examples/cascadeproxy/main.go#L22-L24

they do not seem to arrive at the `endproxy`.

gbraad:
@liudanking WDYT?

liudanking:
I will check it asap

获取 Outlook for iOS<https://aka.ms/o0ukef>

________________________________
收件人: Gerard Braad <notifications@github.com>
发送时间: 星期五, 八月 10, 2018 13:45
收件人: elazarl/goproxy
抄送: Danieliu; Mention
主题: Re: [elazarl/goproxy] HTTP in cascadeproxy example returns Proxy Authentication Required (#306)


@liudanking<https://github.com/liudanking> WDYT?

―

[...]

PR #385, opened 2020-05-02
adresses  #306, adds late handlers in proxy to enable flexible setting o

blunghamer:
auth headers are stripped after request handlers are processed, added outbound handler chain to reset or set new header variables after deleting the inbound ones at the hop, added an example to use, this addresses the headers not getting through to the end proxy in the examples


ErikPelli:
What is the usage for that? Please explain me what could be a case where an additional type of handlers is useful

PR #387, opened 2020-05-14
Avoid generate generate certificate twice.

beppler:
It uses a cache and employs double check pattern to avoid generate twice.

This is almost the same change of #351, but does not have the change of generate a certificate twice on a race condition.

### DIFF
--- a/signer.go
+++ b/signer.go
@@ -15,6 +15,8 @@ import (
 	"net"
 	"runtime"
 	"sort"
+	"strings"
+	"sync"
 	"time"
 )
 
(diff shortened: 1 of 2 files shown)


elazarl:
Thanks, looks good.

How do user control the size of the cache?

beppler:
Now I'm not controlling, but this kind of control can be added.

elazarl:
My concern is, that the memory will blow up after a long run

beppler:
sync.Map does not have a method to count entries.

To make a LRU policy to remove entries it will be necessary to add an additional field to mark the certificate usage and then remove the ones that have older usage date or less usage count.

What do you think about it?

beppler:
Only for record, today I'm using this code to make the cache based on `CertStorage` interface:

```csharp
type certCache struct {
	cache  sync.Map
	mutext sync.Mutex
}

func (storage *certCache) Fetch(hostname string, gen func() (*tls.Certificate, error)) (*tls.Certificate, error) {
	if cachedCert, found := storage.cache.Load(hostname); found {
		cert := cachedCert.(*tls.Certificate)
		return cert, nil
	}

	storage.mutext.Lock()
	defer storage.mutext.Unlock()

	if cachedCert, found := storage.cache.Load(hostname); found {
		cert := cachedCert.(*tls.Certificate)
		return cert, nil
	}

	cert, err := gen()
	if err != nil {
		return nil, fmt.Errorf("Could sign certificate for %s: %v", hostname, err)
	}

	storage.cache.Store(hostname, cert)

	return cert, nil
}
```

It has the same problem.

beppler:
Hi, I included a simple FIFO strategy to keep the number of certificates in an controlled number (4096).

beppler:
Rebased my commits to simplify review.

beppler:
Looking it better, I think that the CertStorage implementation is better in this case, using it I can set the cache site and remove policy in ways that I can not do in the core implementation.

Because of it I'm closing the pull request.

Issue #395, opened 2020-08-03
excuse me, I don't know what this code means ？


telanflow:
https://github.com/elazarl/goproxy/blob/153946a5f232bb80a34d847c6397415b9389407a/proxy.go#L150-L161

`origBody`  ==  `resp.Body` 

Isn't it always the same ?

elazarl:
You can change the body

telanflow:
`body` changed in before and this code is not change it ? Sorry, I still don't understand @elazarl

telanflow:
@elazarl 
I looked at the commit history, it started like this

```go
                origBody := resp.Body
		resp = proxy.filterResponse(resp, ctx)

		ctx.Logf("Copying response to client %v [%d]", resp.Status, resp.StatusCode)
		// http.ResponseWriter will take care of filling the correct response length
		// Setting it now, might impose wrong value, contradicting the actual new
		// body the user returned.
		// We keep the original body to remove the header only if things changed.
		// This will prevent problems with HEAD requests where there's no body, yet,
		// the Content-Length header should be set.
		if origBody != resp.Body {
			resp.Header.Del("Content-Length")
		}
```

elazarl:
@telanflow note that `proxy.filterResponse` can change the body.

If we have a new body, the `Content-Length` header might be wrong, hence we remove it.

Does that make sense?

telanflow:
@elazarl 

I know, but the code is `origBody := resp.body` after `proxy.filterResponse`

https://github.com/elazarl/goproxy/blob/153946a5f232bb80a34d847c6397415b9389407a/proxy.go#L135-L168

It used to be that way

```go
origBody := resp.Body
resp = proxy.filterResponse(resp, ctx)
defer origBody.Close()
ctx.Logf("Copying response to client %v [%d]", resp.Status, resp.StatusCode)
// http.ResponseWriter will take care of filling the correct response length
// Setting it now, might impose wrong value, contradicting the actual new
// body the user returned.
// We keep the original body to remove the header only if things changed.
// This will prevent problems with HEAD requests where there's no body, yet,
// the Content-Length header should be set.
if origBody != resp.Body {
	resp.Header.Del("Content-Length")
}
```

elazarl:
I'm probably missing something.
Can you describe a case where this could be a bug (client send request, goproxy filters it to be X, etc.)/

Thanks

telanflow:
#396

PR #396, opened 2020-08-03
unit testing to response Content-Length. 

telanflow:
#395


elazarl:
That looks great! Thanks!

PR #398, opened 2020-08-09
fix a regular expression in eavesdropper example

### LINKED ISSUE #379: eavesdropper example completely broken
literally just doesn't work if you set it as your browser.

### LINKED ISSUE #386: goproxy.AlwaysReject seems doesn't work
**which code**

this example <https://github.com/elazarl/goproxy/blob/master/examples/goproxy-eavesdropper/main.go#L22>

**what I do**

```
curl https://test.baidu.com/ -x "http://127.0.0.1:8080" -k
```

**what I except**

connect failed

**what actually happen**

connected

image

**some supplement**

in line 22, we got 

```golang
proxy.OnRequest(goproxy.ReqHostMatches(regexp.MustCompile("^.*baidu.com$"))).
		HandleConnect(goproxy.AlwaysReject)
```

in the doc <https://godoc.org/github.com/elazarl/goproxy#ReqProxyConds.HandleConnect> , we find the explain **rejects all CONNECT requests**, so I think my connect should be reject, but I connected successfully, it seems doesn't work

image

Is that my problem?

CoolSpring8:
Fixes #386
Fixes #379

### DIFF
--- a/examples/goproxy-eavesdropper/main.go
+++ b/examples/goproxy-eavesdropper/main.go
@@ -19,7 +19,7 @@ func orPanic(err error) {
 
 func main() {
 	proxy := goproxy.NewProxyHttpServer()
-	proxy.OnRequest(goproxy.ReqHostMatches(regexp.MustCompile("^.*baidu.com$"))).
+	proxy.OnRequest(goproxy.ReqHostMatches(regexp.MustCompile("baidu.*:443$"))).
 		HandleConnect(goproxy.AlwaysReject)
 	proxy.OnRequest(goproxy.ReqHostMatches(regexp.MustCompile("^.*$"))).
 		HandleConnect(goproxy.AlwaysMitm)


CoolSpring8:
This line `"^.*baidu.com$"` was added in 80053202ffae25698d781369a001cdb73794ba07, and was used to block a site's plain HTTP tunneling on a default 80 port. 

1. Solution 1: However, this line doesn't work for now. Not sure what caused this change, but at present there must be an explicit ":80" before "$", which is `"^.*baidu.com:80$"` _**(Solution 1)**_. In this case, it would be better to document that curl -p must be used in companion to test this example.

2. Solution 2: After merging this PR `"baidu.*:443$"` _**(Solution 2)**_, the line would change its meaning and to block HTTPS tunneling on 443 instead (that might be what many people think when they first come to this repo and see the example). In this case, the example won't confuse newcomers ("why this doesn't block the site in my browser?"), but the example loses its uniqueness and would just be the same as the reddit example in readme.

https://github.com/elazarl/goproxy/blob/49ad98f6dac1e021cff612afa5a9d95843efde52/README.md#L139

3. Solution 3: Or something simpler like `"baidu\\.com"` _**(Solution 3)**_ would work for both usage.

@elazarl What's your opinion?

CoolSpring8:
More questions on eavesdropper example: #399

elazarl:
Thanks @CoolSpring8 ! Very important documentation fix. I apologize for the disinformation.

PR #406, opened 2020-10-06
Respond with 200 Ok on successful connection

RoganDawes:
Otherwise the browser will wait forever.


elazarl:
Why do you need to write it yourself? The remote site would respond for you, won't it?

RoganDawes:
A CONNECT request works as follows:

Client:
```
CONNECT target:80 HTTP/1.0

```

Proxy:
```
HTTP/1.0 200 Ok

```
Client:
```
GET / HTTP/1.0
Host: target

```
Target:
```
HTTP/1.0 200 Ok

```

If the proxy doesn't respond with the initial 200 Ok, the client will never issue the subsequent request to the remote site.

PR #412, opened 2020-11-12
Global defaultTLSConfig

lubiedo:
Simple way to make changes to the default config for TLS.

### DIFF
--- a/certs.go
+++ b/certs.go
@@ -21,6 +21,8 @@ var defaultTLSConfig = &tls.Config{
 	InsecureSkipVerify: true,
 }
 
+var DefaultTLSConfig = &defaultTLSConfig
+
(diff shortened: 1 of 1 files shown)


elazarl:
Can you please explain the motivation for the change?

lubiedo:
To make changes to the underlaying `tls.Config{}` structure hold by `defaultTLSConfig`. That configuration is then used by `TLSConfigFromCA` and subsequently OkConnect, MitmConnect, etc. If the default TLS configuration is global we can make changes that will by used by those connect actions (ie. make use of `KeyLogWriter`).

ErikPelli:
If you need to change the TLS configuration, exposing a global variable isn't the solution.
Instead, you can define your own ConnectAction, based on OkConnect, that returns your TLS configuration in the `TLSConfig()` function.

Issue #352, opened 2019-07-15
Can't chain outputs from one handler to another


dmjones:
I was hoping to construct a chain of filters by doing something like this:

```go
proxy := goproxy.NewProxyHttpServer()
proxy.OnRequest().Do(filter1)
proxy.OnRequest().Do(filter2)
```

However, I've noticed filters get passed the original HTTP request, rather than the result from an earlier filter in the chain:

```go
// proxy.go:58
func (proxy *ProxyHttpServer) filterRequest(r *http.Request, ctx *ProxyCtx) (req *http.Request, resp *http.Response) {
	req = r
	for _, h := range proxy.reqHandlers {
		req, resp = h.Handle(r, ctx)   // <-------- r, not req

		if resp != nil {
			break
		}
	}
	return
}
```

Notice how `h.Handle` is invoked on `r` not `req`. Is this intentional or an error?

In the meantime, I can work around this by creating one handler that performs the chaining I need.

elazarl:
IIRC I was thinking that one should concatenate the handler creation, to remove ambiguity.

Do you have a better idea? Would the reader understand the code snippet you wrote better? I do not have particular opinion, you might be right here.

jgroffen:
Hello @elazarl,

I hit the same problem and agree with @dmjones, the reason handlers receive AND return the request object is to give handlers an opportunity to change the request, and have those changes persist. Keep in mind that a common pattern here is to create and return a new request, replacing the one passed in to the handler.

If you pass the original request object (r in this case) for each call to a handler then any changes the handlers make are lost.

This is of course except for the last time a handler called - as only the last return value from h.Handler is retained,

This led to some buggy behavior where once I chained more than one handler I would lose changes to the request made by all but the last handler called - this wasn't easy to diagnose as it appears to work properly when you only apply one handler.

In my example I have two different activities, both of which need to store information in context on the request, then read that information in the response. I did this with a call that looks something like this:

```go
goproxy.OnRequest().DoFunc(requestHandler1)
goproxy.OnRequest().DoFunc(requestHandler2)

func requestHandler1(req *http.Request, ctx *goproxy.ProxyCtx) (req *http.Request, resp *http.Response) {
    # create some context information
    ...
    req = req.WithContext(context.WithValue(req.Context(), ContextKey1, someContextValue))
}

func requestHandler2(req *http.Request, ctx *goproxy.ProxyCtx) (req *http.Request, resp *http.Response) {
    # create some other context information
    ...
[...]

PR #415, opened 2020-12-15
Bugfix in proxy.go while using multiple handlers

chaudhryfaisal:
should pass updated req for remaining handlers


CoolSpring8:
related to #352

ErikPelli:
This change can be useful to add information / modify a request and use the new one in the next filters.
However, I'm not sure if someone is relying on the existent behaviour, where the changes are not persisted.
Could it be useful or I can just assume that if someone modifies the request, he wants to persist the change?
What do you think?

ErikPelli:
I deeply analyzed this change and decided to merge it as it is, because it's the behaviour most users would expect from chaining handlers.
Currently, req is set using the last handler, and it can modify its value.
From this behaviour, I would expect that also intermediary handlers can modify it, but currently it's not.
Users expects the request modification in the latest handler, so we can apply it to the others too.

Issue #419, opened 2021-01-10
examples/goproxy-websockets import ecordell/goproxy instead elazarl/goproxy


4n70w4:
Is this a mistake or is there some justification for this?

If use `ecordell/goproxy` got error:

```
panic: Error parsing builtin CA x509: RSA key missing NULL parameters

goroutine 1 [running]:
github.com/ecordell/goproxy.init.0()
	C:/portable/Go/bin/src/github.com/ecordell/goproxy/certs.go:10 +0x264
```

If use `elazarl/goproxy` got error:

```
2021/01/10 18:23:06 Starting echo server
2021/01/10 18:23:06 Starting proxy server
2021/01/10 18:23:06 dial:proxy: unknown scheme: wss
```

ErikPelli:
https://github.com/elazarl/goproxy/pull/420 resolved this issue

PR #420, opened 2021-01-10
Update main.go

4n70w4:
https://github.com/elazarl/goproxy/issues/419


elazarl:
Probably mistaken taken from a PR. Thanks

Issue #309, opened 2018-08-21
Using AlwaysMitm with Basic auth error EOF


anhhh11:
I want to use both ext/auth and AlwaysMitm for HTTP and HTTPS; It work for HTTP but after enter username, password - HTTPS response show EOF:
If I comment out line (1)  then HTTPS auth work but HTTP doesn't. If keeping (1) then error EOF.

```
	proxy.OnRequest().Do(Basic(realm, func(user, pwd string) bool {
		return user == username && password == pwd
	})) // (1) - For HTTP
	proxy.OnRequest().HandleConnect(BasicConnectWithMITM(realm, func(user, pwd string) bool {
		return user == username && password == pwd
	})) // (2) - For HTTPS

       func BasicConnectWithMITM(realm string, f func(user, passwd string) bool) goproxy.HttpsHandler {
	return goproxy.FuncHttpsHandler(func(host string, ctx *goproxy.ProxyCtx) (*goproxy.ConnectAction, string) {
		if !auth(ctx.Req, f) {
			ctx.Resp = auth.BasicUnauthorized(ctx.Req, realm)
			return goproxy.RejectConnect, host
		}
		return goproxy.MitmConnect, host // <-- I've changed here
	})
}
```

Log
2018/08/21 01:51:41 [177] INFO: req somesite.com:443
2018/08/21 01:51:41 [171] INFO: Exiting on EOF
....

anhhh11:
Not found the root of cause but match proxy.OnRequest Do(proxy_ext.Basic) only on HTTP request seem work:
```
matchUserNamePasswordHandler := proxy_ext.MachUsernamePassword(username, password)
proxy.OnRequest(proxy_ext.IsNotHttps).Do(proxy_ext.Basic(realm, matchUserNamePasswordHandler))
func MachUsernamePassword(username, password string) func(user, pwd string) bool {
	return func(user, pwd string) bool {
		return user == username && password == pwd
	}
}
```

Issue #416, opened 2021-01-04
Don't work cascade with mitm at the same time? Cannot read TLS response from mitm'd server proxyconnect tcp: tls: first record does not look like a TLS handshake


4n70w4:
Hi! I get code from https://github.com/elazarl/goproxy/blob/0581fc3aee2d07555835bed1a876aca196a4a511/examples/cascadeproxy/main.go

and add Mitm handler to `middleProxy`

```go
middleProxy.OnRequest(goproxy.ReqHostMatches(regexp.MustCompile("^.*$"))).HandleConnect(goproxy.AlwaysMitm)
```

Now all requests failed: Cannot read TLS response from mitm'd server proxyconnect tcp: tls: first record does not look like a TLS handshake

```
2021/01/04 03:37:52 serving end proxy server
2021/01/04 03:37:52 serving middle proxy server
2021/01/04 03:37:53 [001] INFO: Running 1 CONNECT handlers
2021/01/04 03:37:53 [001] INFO: on 0th handler: &{2 <nil> 0x792400} webhook.site:443
2021/01/04 03:37:53 [001] INFO: Assuming CONNECT is TLS, mitm proxying it
2021/01/04 03:37:53 [001] INFO: signing for webhook.site
2021/01/04 03:37:53 [002] INFO: req google.site:443
2021/01/04 03:37:55 [002] INFO: Sending request GET https://google.site:443/
2021/01/04 03:37:56 [002] WARN: Cannot read TLS response from mitm'd server proxyconnect tcp: tls: first record does not look like a TLS handshake
```

or

```
2021/01/04 03:45:09 [002] WARN: Cannot read TLS response from mitm'd server Proxy Authentication Required
```

Is there any way to use Mitm with cascade proxy?

May be related: https://github.com/elazarl/goproxy/issues/177

ngoyal16:
@4n70w4 can you help me to setup MITM i am not able to setup that

PR #428, opened 2021-03-06
Basic auth fixed

clburlison:
**Issue:** When using AlwaysMitm ext/auth/basic.go would attempt to authenticate requests twice resulting in the 407 unauthorized message.

This fixes two outstanding issues with the basic auth behavior of this library.

1. BasicConnect handler was preventing other handlers from running (AlwaysMitm)
2. Basic was trying to authenticate connection that was already authenticated by BasicConnect

Resolves: https://github.com/elazarl/goproxy/issues/309, https://github.com/elazarl/goproxy/issues/416

Below is a simple test case to validate the behavior before/after the patch:

```
package main

import (
	"flag"
	"log"
	"net"
	"net/http"
	"regexp"

	"github.com/elazarl/goproxy"
	"github.com/elazarl/goproxy/ext/auth"
)

func handleRequest(req *http.Request, ctx *goproxy.ProxyCtx) (*http.Request, *http.Response) {
	ip, _, err := net.SplitHostPort(req.RemoteAddr)
	if err != nil {
		log.Print(err)
	}

	log.Printf("[%d] %s --> %s %s", ctx.Session, ip, req.Method, req.URL)

	return req, nil
}

func main() {
	verbose := flag.Bool("v", false, "should every proxy request be logged to stdout")
	addr := flag.String("addr", ":8080", "proxy listen address")
	flag.Parse()
	username, password := "foo", "bar"
	hostmatch := "google.com|neverssl.com|apache.org"
	auth.ProxyBasic(proxy, "Auth", func(user, passwd string) bool {
		return user == username && password == passwd
	})
	proxy.OnRequest(goproxy.ReqHostMatches(regexp.MustCompile(hostmatch))).
		HandleConnect(goproxy.AlwaysMitm)
	proxy.OnRequest().DoFunc(handleRequest)
	proxy.Verbose = *verbose
	log.Fatal(http.ListenAndServe(*addr, proxy))
}
```

### DIFF
--- a/ctx.go
+++ b/ctx.go
@@ -23,6 +23,8 @@ type ProxyCtx struct {
 	Session   int64
 	certStore CertStorage
 	Proxy     *ProxyHttpServer
+	// Will prevent second authentication on the already authenticated requests
+	Authenticated bool
 }
 
 type RoundTripper interface {
--- a/ext/auth/basic.go
+++ b/ext/auth/basic.go
@@ -52,9 +52,10 @@ func auth(req *http.Request, f func(user, passwd string) bool) bool {
 // You probably want to use auth.ProxyBasic(proxy) to enable authentication for all proxy activities
 func Basic(realm string, f func(user, passwd string) bool) goproxy.ReqHandler {
 	return goproxy.FuncReqHandler(func(req *http.Request, ctx *goproxy.ProxyCtx) (*http.Request, *http.Response) {
-		if !auth(req, f) {
+		if !ctx.Authenticated && !auth(req, f) {
 			return nil, BasicUnauthorized(req, realm)
 		}
+		ctx.Authenticated = true
 		return req, nil
 	})
 }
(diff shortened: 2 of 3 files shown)


bynf:
@elazarl Could you merge this PR?

elazarl:
I'm very careful about adding yet another field to the context. Can we do it in some other way?

bynf:
Actually I did't think another way :) But this PR working correctly for me.

bynf:
Sorry, I had a problem after i implemented it. 
'ctx.Authenticated' not returns 'true' for http connect methods. 

I am getting 407 (proxy auth required) for this code. Is there something I did wrong?

`addr := flag.String("addr", ":8080", "proxy listen address")
	hostmatch := flag.String("hostmatch", "^.*$", "hosts to trace (regexp pattern)")
	verbose := flag.Bool("v", true, "verbose output")
	flag.Parse()

	log.SetFlags(log.Lmicroseconds)

	auth.ProxyBasic(proxy, "Auth", func(user, passwd string) bool {
		return true
	})

	proxy.OnRequest(goproxy.ReqHostMatches(regexp.MustCompile(*hostmatch))).
		HandleConnect(goproxy.AlwaysMitm)

	proxy.OnRequest().DoFunc(handleRequest)

	proxy.Verbose = *verbose
	log.Fatal(http.ListenAndServe(*addr, proxy))`

thalesfsp:
@elazarl Given the importance of the MR, could you do the change you are suggesting? Best regards!

guanyufen123:
这个啥时候能修复呢

thalesfsp:
@guanyufen123 just fork and merge the change. This is what happens when the community needs and demand things but authors are reluctant to listen.

elazarl:
@thalesfsp can you please explain me again, why using the `User` pointer is not a good solution in this case?

I'm very open to listen, but was not able to spurr discussion.

clburlison:
Closing . Someone else from the community is welcome to pick up this request if desired.

elazarl:
Thanks @clburlison , I think fork is the best solution here. Me and @ErikPelli try to accomodate for most people's needs, and changes to the core must be done carefully on the main project.

Issue #431, opened 2021-03-17
The dynamic cert date range is too long


xujinzheng:
The dynamic cert date range is too long, at some mobile device will trigger the error as follow:

`ERR_CERT_VALIDITY_TOO_LONG`

https://support.servertastic.com/knowledgebase/article/neterr_cert_validity_too_long-the-server-certificate-has-a-validity-period-that-is-too-long

PR #432, opened 2021-03-17
fix issue: #431 (ERR_CERT_VALIDITY_TOO_LONG)

xujinzheng:
I fixed the cert date range  as follow:

```go
start := time.Unix(time.Now().Unix()-2592000, 0) // 2592000  = 30 day
end := time.Unix(time.Now().Unix()+31536000, 0)  // 31536000 = 365 day
```


titanhw:
I also encountered this problem. I think this pr should be accepted.
@elazarl

PR #435, opened 2021-03-24
Adds a error handler to ConnectAction

mcfedr:
This allows the users to client handle connect errors related to certificates, useful, as in the example, when some clients are using certificate pinning to allow them to connect without mitm


ErikPelli:
This change is too specific and doesn't add much useful behaviour, since `ConnectionErrHandler` already handle this kind of errors.
I'm closing this pull request.

PR #445, opened 2021-07-15
feat: Export Handlers (req, rest, https)

guilhem:
This is usufull when handlers will be dynamicaly created and removed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,14 +109,14 @@ There are 3 kinds of useful handlers to manipulate the behavior, as follows:
 ```go
 // handler called after receiving HTTP CONNECT from the client, and before proxy establish connection 
 // with destination host
-httpsHandlers   []HttpsHandler
+HttpsHandlers   []HttpsHandler
     
 // handler called before proxy send HTTP request to destination host
-reqHandlers     []ReqHandler 
+ReqHandlers     []ReqHandler 
     
 // handler called after proxy receives HTTP Response from destination host, and before proxy forward 
 // the Response to the client.
-respHandlers    []RespHandler 
+RespHandlers    []RespHandler 
 ```
 
 Depending on what you want to manipulate, the ways to add handlers to each handler list are:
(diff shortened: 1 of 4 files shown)


elazarl:
Can you please explain why you can't use the existing methods to manipulate the handlers?

Making a private field public is risky, as it cannot be undone.

guilhem:
@elazarl in current implementation, Handlers are manipulated by `.Do(`, `.HijackConnect(`, `.HandleConnect(` and only with an `= append(`.
It makes impossible to remove elements or to know what are elements inside.

What I need is to Add / List and Remove handlers to be able to do it dynamically (at run-time).

guilhem:
@elazarl an usage example: https://github.com/guilhem/egress-proxy-operator/blob/master/controllers/request_controller.go#L126

elazarl:
I don't see why you need indirection. Making them public should be good enough.

Slice is already a pointer.

Also not sure how you handle concurrency.

guilhem:
> I don't see why you need indirection. Making them public should be good enough.
> 
> Slice is already a pointer.
> 
> Also not sure how you handle concurrency.

I totally agree with you.
But I faced a lot of problems with slices and make it working with pointer. 🤷‍♂️
I need to do more tests.

I also find this article, not sure if applicable: https://medium.com/swlh/golang-tips-why-pointers-to-slices-are-useful-and-how-ignoring-them-can-lead-to-tricky-bugs-cac90f72e77b

guilhem:
@elazarl I did more tests and slice is not updated if it's not a pointer.
Having a pointer also have some interest, it's possible to update content atomically. Just create a new slice with right content and update pointer in 1 step.

elazarl:
@guilhem what I suggest is, add a single handler that has internal handlers, and handle concurrency inside

We'll add this as an extension.

What do you think?

guilhem:
> @guilhem what I suggest is, add a single handler that has internal handlers, and handle concurrency inside
> 
> We'll add this as an extension.
> 
> What do you think?

I will try to do this solution :)

PR #447, opened 2021-08-01
Fix method to deny connections in docs.

areading314:
Saw that RejectConnect wasn't working, but AlwaysReject behaved as expected.


elazarl:
Thanks!

PR #448, opened 2021-08-22
httpError: write error string as body in 502 response

omribahumi:
When `CONNECT` fails, other than getting a 502 response, there's no indication of the failure reason.

This pull request will:
1. Print the dial failure reason to the Go log output
2. Return the error in the 502 response body

I couldn't test this with curl as it suppresses the body for 5xx responses for some reason. Not sure if this is because this is a 502 in response to the CONNECT call.

Before:
```
% echo $'CONNECT nodns.boo HTTP/1.1\r\n' | nc localhost 8080
HTTP/1.1 502 Bad Gateway

```

After:
```
% echo $'CONNECT nodns.boo HTTP/1.1\r\n' | nc localhost 8080
HTTP/1.1 502 Bad Gateway
Content-Type: text/plain
Content-Length: 40

dial tcp: lookup nodns.boo: no such host
```

### DIFF
--- a/https.go
+++ b/https.go
@@ -4,6 +4,7 @@ import (
 	"bufio"
 	"crypto/tls"
 	"errors"
+	"fmt"
 	"io"
 	"io/ioutil"
 	"net"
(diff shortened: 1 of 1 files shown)


nkns165:
@elazarl It looks like a good change, why don't you merge it? I'd love to have this warning log.

elazarl:
@nkns165 I'm worried it'd log too much for some people, maybe we need a better tunable logs.

nkns165:
Thank you so much...!

PR #456, opened 2021-12-04
Improve certstorage example

tetafro:
I've added more info and one more example to `CertStore` example.

I actually just looked through these discussions and copied useful information, so all credits to people from there:

- https://github.com/elazarl/goproxy/pull/332
- https://github.com/elazarl/goproxy/pull/330
- https://github.com/elazarl/goproxy/pull/284

### DIFF
--- a/examples/goproxy-certstorage/README.md
+++ b/examples/goproxy-certstorage/README.md
@@ -1,5 +1,13 @@
-# certstorage
+# In-memory storage for temporary certificates
 
-## use certstorage 
+By default goproxy will generate TLS certificate for every request. Generating
+certificates is computationally expensive, so this leads to performance issues
+when there are many requests to the same host.
 
-## this could make https proxy faster
+Certificates storage allows the reuse of certificates.
+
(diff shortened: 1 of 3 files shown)


ErikPelli:
Having a mutx for every single host looks like an anti-pattern to me.
This will also cause memory leaks if there are a lot of different hosts, since the older one are never cleaned uo.

tetafro:
I don't really remember my thinking from 4 years ago, but looks like it's just the most straight-forward example I could come up with.
This definitely can be improved in real life, using some caching with eviction policy instead of a map.

ErikPelli:
Uhm ok, do you think it should still be merged or the current simple example is enough as a simple example for someone that never used this library? (Do you still use it?)

tetafro:
Wait, I just noticed that you did pretty much the same in the [current example](https://github.com/elazarl/goproxy/blob/master/examples/certstorage/cache.go). Why are we discussing this PR then? Let's just close it.

PR #460, opened 2021-12-28
fix: honour KeepHeader on https connections

### LINKED ISSUE #453: The keepHeaders option does not work on https connections
image
https://github.com/elazarl/goproxy/blob/master/https.go#L231

As this image show, it doesn't check whether keepHeaders enable or not before call `removeProxyHeaders`.

mvanholsteijn:
My https proxy inserts a new proxy-authorization header in the OnRequest() method, which is removed in https.go

- I added the if !proxy.KeepHeader condition around the removeProxyHeaders call in https.go just like it was added in proxy.go
- I changed the name from removeProxyHeaders to RemoveProxyHeaders to allow the headers to be removed from an OnRequest() method. 

closes #453


ErikPelli:
Can you resolve the conflicts?

PR #462, opened 2022-01-15
Improve an example of transparent proxy: use DialContext() instead of Dial()

hgsgtk:
First, I gratefully appreciate your easy-to-understand sample code!
It is very useful for me to gain knowledge to support HTTPS.

I'm afraid we can improve this example by avoiding using the deprecated function `Dial()`.
Other minor text corrections are included.

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,7 +97,7 @@ a `ReqCondition` accepting only requests directed to "www.reddit.com".
 
 `DoFunc` will receive a function that will preprocess the request. We can change the request, or
 return a response. If the time is between 8:00am and 17:00pm, we will reject the request, and
-return a precanned text response saying "do not waste your time".
+return a pre-canned text response saying "do not waste your time".
 
 See additional examples in the examples directory.
 
--- a/examples/goproxy-transparent/README.md
+++ b/examples/goproxy-transparent/README.md
@@ -1,10 +1,10 @@
 # Transparent Proxy
 
-This transparent example in goproxy is meant to show how to transparenty proxy and hijack all http and https connections while doing a man-in-the-middle to the TLS session.  It requires that goproxy sees all the packets traversing out to the internet.  Linux iptables rules deal with changing the source/destination IPs to act transparently, but you do need to setup your network configuration so that goproxy is a mandatory stop on the outgoing route.  Primarily you can do this by placing the proxy inline.  goproxy does not have any WCCP support itself; patches welcome.
+This transparent example in goproxy is meant to show how to transparent proxy and hijack all http and https connections while doing a man-in-the-middle to the TLS session.  It requires that goproxy sees all the packets traversing out to the internet.  Linux iptables rules deal with changing the source/destination IPs to act transparently, but you do need to set up your network configuration so that goproxy is a mandatory stop on the outgoing route.  Primarily you can do this by placing the proxy inline.  goproxy does not have any WCCP support itself; patches are welcome.
 
 ## Why not explicit?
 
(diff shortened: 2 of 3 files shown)


examples/goproxy-transparent/README.md
```diff
@@ -1,10 +1,10 @@
...
 
 ## Why not explicit?
 
-Transparent proxies are more difficult to maintain and setup from a server side, but they require no configuration on the client(s) which could be in unmanaged systems or systems that don't support a proxy configuration.  See the [eavesdropper example](https://github.com/elazarl/goproxy/blob/master/examples/goproxy-eavesdropper/main.go) if you want to see an explicit proxy example.
+Transparent proxies are more difficult to maintain and set up from a server side, but they require no configuration on the client(s) which could be in unmanaged systems or systems that don't support a proxy configuration.  See the [eavesdropper example](https://github.com/elazarl/goproxy/blob/master/examples/goproxy-eavesdropper/main.go) if you want to see an explicit proxy example.
```
hgsgtk, line 7: [FYI]

It appears we should use two words "set up" when it functions as a verb.

https://grammarist.com/spelling/set-up-vs-setup/

> Setup is one word when it is a noun (e.g., “it was a setup!”) or an adjective (e.g., “follow the setup instructions”). It is two words—set up—when it functions as a verb (e.g., “I’m going to set up the computer”).

README.md
```diff
@@ -97,7 +97,7 @@ a `ReqCondition` accepting only requests directed to "www.reddit.com".
 
 `DoFunc` will receive a function that will preprocess the request. We can change the request, or
 return a response. If the time is between 8:00am and 17:00pm, we will reject the request, and
-return a precanned text response saying "do not waste your time".
+return a pre-canned text response saying "do not waste your time".
```
hgsgtk, line 100: [FYI] I googled and pre-canned seems to be a better way to convey it.

https://www.google.com/search?q=precanned&oq=precanned&aqs=chrome.0.69i59.187j0j4&sourceid=chrome&ie=UTF-8

examples/goproxy-transparent/transparent.go
```diff
@@ -110,17 +112,17 @@ func main() {
 }
 
 // copied/converted from https.go
-func dial(proxy *goproxy.ProxyHttpServer, network, addr string) (c net.Conn, err error) {
-	if proxy.Tr.Dial != nil {
-		return proxy.Tr.Dial(network, addr)
+func dial(ctx context.Context, proxy *goproxy.ProxyHttpServer, network, addr string) (c net.Conn, err error) {
+	if proxy.Tr.DialContext != nil {
```
hgsgtk, line 116: [FYI] I modified it to use Transport.DialContext.

https://pkg.go.dev/net/http#Transport

```
	// DialContext specifies the dial function for creating unencrypted TCP connections.
	// If DialContext is nil (and the deprecated Dial below is also nil),
	// then the transport dials using package net.
	//
	// DialContext runs concurrently with calls to RoundTrip.
	// A RoundTrip call that initiates a dial may end up using
	// a connection dialed previously when the earlier connection
	// becomes idle before the later DialContext completes.
	DialContext func(ctx context.Context, network, addr string) (net.Conn, error)

	// Dial specifies the dial function for creating unencrypted TCP connections.
	//
	// Dial runs concurrently with calls to RoundTrip.
	// A RoundTrip call that initiates a dial may end up using
	// a connection dialed previously when the earlier connection
	// becomes idle before the later Dial completes.
	//
	// Deprecated: Use DialContext instead, which allows the transport
	// to cancel dials as soon as they are no longer needed.
	// If both are set, DialContext takes priority.
	Dial func(network, addr string) (net.Conn, error)
```

elazarl:
Thanks!

PR #463, opened 2022-01-21
allow goproxy to operate like a front proxy

sgargan:
in this mode the NonProxyHandler is skipped and it act very much like nginx/envoy, proxying http requests without proxy headers.

### DIFF
--- a/proxy.go
+++ b/proxy.go
@@ -22,10 +22,12 @@ type ProxyHttpServer struct {
 	Verbose         bool
 	Logger          Logger
 	NonproxyHandler http.Handler
-	reqHandlers     []ReqHandler
-	respHandlers    []RespHandler
-	httpsHandlers   []HttpsHandler
-	Tr              *http.Transport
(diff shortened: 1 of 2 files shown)


elazarl:
Thanks!

Can you please explain the use case?

sgargan:
I'm using this as a proxy embedded in a cli app I'm building, cli starts the proxy and also a browser which is directed to the browser. I'd like the browser to be able to call this like it was a front proxy, without need of proxy config. This will let me make use of the powerful request routing. If I use the non proxy handler, it skips the routing; setting the flag lets me skip the NonProxyHandler and use the routing without needing proxy config. This is great kit btw kudos!

elazarl:
What's the problem with just setting the `NonProxyHandler` to do whatever you want as a web server? That's the intent of it.

sgargan:
If I use the NonProxyHandler i don't get to use any of your powerful route handling, only basic request processing. Once the NonProxyHandler runs it returns and the route handling gets skipped.

PR #466, opened 2022-03-16
Fix the signer signing an invalid certificate

riijj:
The Signer will sign an invalid certificate when hostname is an IPV6 address

E.g :
Assuming the URL is https://[2606:4700:4700::1111]:443 , the CommonName of signed certificate will be: [2606 , the certificate is invalid.

this PR will fix the bug.

### DIFF
--- a/https.go
+++ b/https.go
@@ -47,9 +47,25 @@ type ConnectAction struct {
 }
 
 func stripPort(s string) string {
-	ix := strings.IndexRune(s, ':')
-	if ix == -1 {
-		return s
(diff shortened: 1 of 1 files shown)


elazarl:
Sounds good, isn't there something in the stdlib to do that?

riijj:
> Sounds good, isn't there something in the stdlib to do that?

Forgive me , I'm not sure what you mean

elazarl:
@riijj a function to parse URL and take IPv6 into account

PR #472, opened 2022-04-16
fix: return `Connection established` instead `OK`

hgsgtk:
This commit changes to return `200 Connection established` instead `200 OK`.

```
   Example of a response:

         HTTP/1.0 200 Connection established
         Proxy-agent: Netscape-Proxy/1.1

         ...data tunnelled from the server...
```

ref: https://datatracker.ietf.org/doc/html/draft-luotonen-web-proxy-tunneling-01#section-3.2


elazarl:
Thanks!

PR #473, opened 2022-04-16
Preserve Content-Length in header of HEAD response

ttt43ttt:
There was another PR https://github.com/elazarl/goproxy/pull/339 fixing this issue.
This PR fixes it with a smaller change.


elazarl:
Thanks! Will merge after tests pass

PR #477, opened 2022-05-27
Resolves #475

### LINKED ISSUE #475: goproxy buffers SSE (content-type: text/event-stream)
The issue is well-described in https://github.com/coreos/goproxy/issues/5.

When using goproxy with SSE, the events are buffered, and a client just keeps waiting for data over an open TCP connection.

Scenario:

- set up an event source for server-side events, e.g. like in https://jasonbutz.info/2018/08/server-sent-events-with-node/
- curl it through goproxy (or use a proper SSE client in javascript using an EventSource)

Problem:
- goproxy buffers server-side events instead of proxying them as they are received

waggledans:
Flush response body for server-side events to avoid events buffering by goproxy.
I validated that it works with sse.

```
$ curl -x localhost:8080 localhost:3000/events
retry: 2000
id: 1
event: data
data: {"id":1,"data":"287039","time":"8:15:06 PM","final":false}
id: 2
event: data
data: {"id":2,"data":"267182","time":"8:15:07 PM","final":false}
id: 3
event: data
data: {"id":3,"data":"657946","time":"8:15:07 PM","final":false}
```

### DIFF
--- a/proxy.go
+++ b/proxy.go
@@ -107,6 +107,20 @@ func removeProxyHeaders(ctx *ProxyCtx, r *http.Request) {
 	r.Header.Del("Connection")
 }
 
+type flushWriter struct {
+	w io.Writer
+}
+
+func (fw flushWriter) Write(p []byte) (int, error) {
+	n, err := fw.w.Write(p)
+	if f, ok := fw.w.(http.Flusher); ok {
+		// only flush if the Writer implements the Flusher interface.
+		f.Flush()
+	}
+
+	return n, err
+}
+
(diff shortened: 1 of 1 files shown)


waggledans:
I could add a test that starts a server and fetches events via the proxy but it would require updating dependencies in go.mod (and, ideally, setting go version to 1.17 to leverage [modules graph pruning](https://go.dev/ref/mod#graph-pruning) to avoid pulling go-sse to other packages that use goproxy). wdyt @elazarl ?

elazarl:
@waggledans sounds good, I'll be happy to approve the PR

PR #484, opened 2022-09-07
fix: proxy.ConnectDial with URLs which require no-proxy

thanhkaiba:
In system like: 

User Computer ---> GoProxy (elazarl proxy) ---> Second Proxy ---> Network

Currently, all URLs pass to ConnectAccept is sent to Second Proxy but some URLs need send directly to network by NO_PROXY config. So we need add check for URLs require no_proxy


ErikPelli:
This can be configured in the proxy configuration without need to add this exception inside the core library,
Have a look at `[]HttpsHandler` proxy handlers to set this exception.

PR #488, opened 2022-11-07
IsLocalHost: support host:port

### LINKED ISSUE #487: IsLocalHost only works with default ports
It was originally observed in https://github.com/elazarl/goproxy/issues/289.

The `IsLocalHost` function uses `req.URL.Host` https://github.com/elazarl/goproxy/blob/a0805db90819b464f8fadc660e23a6a6b9d4942f/dispatcher.go#L105

It can be [host or host:port](https://cs.opensource.google/go/go/+/refs/tags/go1.19.3:src/net/url/url.go;l=363).
In the latter case the check simply does not work. 
It's not documented and it simply looks like a bug. 
It should be using the [Hostname](https://cs.opensource.google/go/go/+/refs/tags/go1.19.3:src/net/url/url.go;bpv=1;bpt=1;l=1143?gsn=Hostname&gs=kythe%3A%2F%2Fgo.googlesource.com%2Fgo%3Flang%3Dgo%3Fpath%3Dnet%2Furl%23method%2520URL.Hostname) function instead.

mmatczuk:
Without this patch IsLocalHost does not work for URLs with port specified i.e. it works for `http://localhost` but does not work for `http://localhost:80` or `http://localhost:10000`.

Fixes #487

### DIFF
--- a/dispatcher.go
+++ b/dispatcher.go
@@ -97,15 +97,22 @@ func ReqHostIs(hosts ...string) ReqConditionFunc {
 	}
 }
 
-var localHostIpv4 = regexp.MustCompile(`127\.0\.0\.\d+`)
-
-// IsLocalHost checks whether the destination host is explicitly local host
-// (buggy, there can be IPv6 addresses it doesn't catch)
+// IsLocalHost checks whether the destination host is localhost.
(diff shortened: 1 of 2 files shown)


ErikPelli:
LGTM

PR #489, opened 2022-11-08
proxy: return 502 Bad Gateway on round-trip error

### LINKED ISSUE #51: Status code on unresolvable hosts.
If we try to access through goproxy host which doesn't have DNS record goproxy returns 500 (Internal Server Error) HTTP status code, while squid-cache returns 502 (Bad Gateway) HTTP status code. Shouldn't goproxy simulate squid's behavior?

```
$ curl -v --proxy 127.0.0.1:3129  --url http://no.exists/
* About to connect() to proxy 127.0.0.1 port 3129 (#0)
*   Trying 127.0.0.1...
* Adding handle: conn: 0x1d71bd0
* Adding handle: send: 0
* Adding handle: recv: 0
* Curl_addHandleToPipeline: length: 1
* - Conn 0 (0x1d71bd0) send_pipe: 1, recv_pipe: 0
* Connected to 127.0.0.1 (127.0.0.1) port 3129 (#0)
> GET http://no.exists/ HTTP/1.1
> User-Agent: curl/7.32.0
> Host: no.exists
> Accept: */*
> Proxy-Connection: Keep-Alive
> 
< HTTP/1.1 500 Internal Server Error
< Content-Type: text/plain; charset=utf-8
< Date: Tue, 20 May 2014 13:13:02 GMT
< Content-Length: 41
< 
dial tcp: lookup no.exists: no such host
* Connection #0 to host 127.0.0.1 left intact
$ 
```

mmatczuk:
Currently, it's not possible to distinguish target server internal error from using a wrong server or server being down. This patch changes status code 500 to 502 in case we cannot get a response from upstream.

Below snippets from the HTTP specification:

> The HyperText Transfer Protocol (HTTP) 502 Bad Gateway
> server error response code indicates that the server, while acting as a gateway or proxy, received an invalid response from the upstream server.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/502

vs.

> The HyperText Transfer Protocol (HTTP) 500 Internal Server Error
> server error response code indicates that the server encountered an unexpected condition that prevented it from fulfilling the request.
> This error response is a generic "catch-all" response. Usually, this indicates the server cannot find a better 5xx error code to response. Sometimes, server administrators log error responses like the 500 status code with more details about the request to prevent the error from happening again in the future.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/500

Fixes #51

### DIFF
--- a/proxy.go
+++ b/proxy.go
@@ -168,16 +168,29 @@ func (proxy *ProxyHttpServer) ServeHTTP(w http.ResponseWriter, r *http.Request) 
 		resp = proxy.filterResponse(resp, ctx)
 
 		if resp == nil {
-			var errorString string
-			if ctx.Error != nil {
-				errorString = "error read response " + r.URL.Host + " : " + ctx.Error.Error()
-				ctx.Logf(errorString)
-				http.Error(w, ctx.Error.Error(), 502)
+			var (
+				errorString string
+				code        int
+			)
+
+			if e, ok := ctx.Error.(net.Error); ok {
+				if e.Timeout() {
+					errorString = "timed out connecting to remote host: " + e.Error()
+					code = http.StatusGatewayTimeout
+				} else {
+					errorString = "failed to connect to remote host: " + e.Error()
+					code = http.StatusBadGateway
+				}
+			} else if ctx.Error != nil {
+				errorString = "failed to read response from remote host: " + e.Error()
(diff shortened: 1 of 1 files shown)


mmatczuk:
I'm closing this PR, I'd open a new one with a better implementation.

PR #498, opened 2023-02-21
connection time TLS SNI support

mixcode:
TLSConfigFromCA() changed to return a config with run-time signing function to properly support TLS Server Name Indication.
Also, a small glitch regarding host name string is fixed.


ErikPelli:
Why should we overwrite the host (which is obtained using r.URL.Host) with the ClientHello value?
We already know the host, this seems not useful to me.
Is there any specific case that you are trying to solve?

PR #503, opened 2023-07-06
adding support for deprecated SPDY/3.1 websocket upgrades

RomanManz:
Not much to say about this one :-)
Just to support clients that still use SPDY/3.1.

### DIFF
--- a/websocket.go
+++ b/websocket.go
@@ -22,7 +22,8 @@ func headerContains(header http.Header, name string, value string) bool {
 
 func isWebSocketRequest(r *http.Request) bool {
 	return headerContains(r.Header, "Connection", "upgrade") &&
-		headerContains(r.Header, "Upgrade", "websocket")
+		( headerContains(r.Header, "Upgrade", "websocket") ||
+			headerContains(r.Header, "Upgrade", "SPDY/3.1"))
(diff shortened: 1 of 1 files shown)


ErikPelli:
Is this still needed? SPDY looks like a deprecated protocol and I think you should keep your own fork to support it.
By the way, this Pull request would allow only one specific version of it (v3.1), I don't think it's ok, since there are also other versions.

elazarl:
@RomanManz can you be more specific about the client? I don't think these are relevant today.

RomanManz:
Hello @elazarl,
I've just checked, the latest versions of the Openshift cli called 'oc' still uses SPDY/3.1 (kubectl itself doesn't use SPDY though). Would that be reason enough to take it? Should the PR be adjusted to support other SPDY versions as well?
Otherwise I am happy to close and keep it private.
Thanks!

ErikPelli:
We changed the code in the master branch to specifically use WebSocket support from the Go standard library.
I don't think the change proposed in the current PR is still possible, since Go stdlib doesn't support this SPDY protocol.

RomanManz:
Thanks a lot for the feedback, closing the PR now.

PR #504, opened 2023-07-06
Connect related changes - add optional hooks

RomanManz:
Commit b9ddce5b948fb229126df449803ff0efbe657d66 adding return value to connectReqHandler
* allows the connectReqHandler to return an error in order to interrupt requests when necessary
Commit fff8ef11b152ec436838d748a1796424bcb5fe20 adding optional connectRespHandler
* introduces optional support for a connectResponseHandler
* allows to interrupt a connection - our use case was to check for specific headers returned from the chained proxy
Commit f0e694013577123774066015f8819a6e8b136bd5 calling Hijack in handleHttps if defined
* adds hijack support to the connectMitm and connectHTTPMitm actions
* our use case was to inject additional headers in the connect responses when needed


elazarl:
The use case is, you have a CA that singed another certificate once, and you want to use this certificate to sign websites, right?

Can you please add unit tests for the certificate functions?

ErikPelli:
What is the purpose of this Pull Request? Can you explain the problem that you are trying to solve?

RomanManz:
> The use case is, you have a CA that singed another certificate once, and you want to use this certificate to sign websites, right?
> 
> Can you please add unit tests for the certificate functions?

Yes @elazarl , happy to add the unit tests.

RomanManz:
Hello @ErikPelli ,
the concrete use case was to add an authentication flow at proxy level using 'Proxy-Authorization', and to do it in two steps:
client -> proxy 1 resposible for the auth -> proxy 2 in front of service -> service
The commits relate to those phases of the auth flow:
1. adding return value to connectReqHandler
    => allow proxy 1 reqhandler to perform auth and to be able to fail
2. adding optional connectRespHandler
    => in case of chained proxies, allow proxy 1 to intercept replies from proxy 2, for ex. auth rejected/expired
3. calling Hijack in handleHttps if defined
    => allow proxy 2 to check Proxy-Authorization header and return a custom auth rejected/expired if needed
This is very specific, but may still help others.
Otherwise I am happy to keep it local.
Thanks.

https.go
```diff
@@ -533,10 +545,18 @@ func (proxy *ProxyHttpServer) NewConnectDialToProxy(httpsProxy string) func(netw
 }
 
 func (proxy *ProxyHttpServer) NewConnectDialToProxyWithHandler(
-	httpsProxy string,
-	connectReqHandler func(req *http.Request),
+  https_proxy string,
+  connectReqHandler func(req *http.Request) error,
+) func(network, addr string) (net.Conn, error) {
+	return proxy.NewConnectDialToProxyWithMoreHandlers(https_proxy, nil, nil)
```
ErikPelli, line 551: Here you are passing nil as the connectReqHandler, can you instead pass the input callback?

https.go
```diff
@@ -533,10 +545,18 @@ func (proxy *ProxyHttpServer) NewConnectDialToProxy(httpsProxy string) func(netw
...
+}
+
+func (proxy *ProxyHttpServer) NewConnectDialToProxyWithMoreHandlers(
+  https_proxy string,
```
ErikPelli, line 555: Please avoid snake case for the input variable names, keep it consistent with the codebase

ErikPelli:
@RomanManz can you fix the linting issues reported by the pipeline?

https.go
RomanManz: Sorry, this happened when I fixed the merge conflicts, fixed now, thanks for pointing this out!

https.go
RomanManz: Fixed.

ErikPelli:
There is still a linting error, I think it's because you aren't using the correct number of spaces for the function parameters

PR #505, opened 2023-07-06
Handle https to http override in TLS MitM websocket connection

### LINKED ISSUE #506: Cannot connect HTTP websocket when performing TLS MitM
In a typical MitM over CONNECT scenario there is possibility to override request schema from secure to insecure. Example: when using TLS MitM we can redirect https service to a local http server by request URL manipulation using OnRequest handler.
But when there is a try to open websocket connection, there is an error - always secure TLS connection is enforced regardless to request URL schema.
Fix in #505

dlski:
In a typical MitM over CONNECT scenario there is possibility to override request schema from secure to insecure. Example: when using TLS MitM we can substitute https service using a local http server by manipulating request URL schema in OnRequest handler.
But when there is a try to open websocket connection, there is an error - always secure TLS connection is enforced regardless to request URL schema. This PR fixes that.
Resolves #506

### DIFF
--- a/https.go
+++ b/https.go
@@ -242,7 +242,12 @@ func (proxy *ProxyHttpServer) handleHttps(w http.ResponseWriter, r *http.Request
 				if resp == nil {
 					if isWebSocketRequest(req) {
 						ctx.Logf("Request looks like websocket upgrade.")
-						proxy.serveWebsocketTLS(ctx, w, req, tlsConfig, rawClientTls)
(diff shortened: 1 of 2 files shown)


LukaSikic:
@elazarl can you please review and hopefully merge? I face same issue as described in #506 and this PR fixes it

elazarl:
Apologies, missed that

PR #507, opened 2023-07-13
Added control for the nil request

### LINKED ISSUE #502: panic: runtime error: invalid memory address or nil pointer dereference
Similar to [this](https://github.com/elazarl/goproxy/issues/461)

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x38 pc=0x6b3260]

goroutine 1287 [running]:
github.com/elazarl/goproxy.(*ProxyHttpServer).handleHttps.func2()
        /go/pkg/mod/github.com/elazarl/goproxy@v0.0.0-20221015165544-a0805db90819/https.go:249 +0xd60
created by github.com/elazarl/goproxy.(*ProxyHttpServer).handleHttps
        /go/pkg/mod/github.com/elazarl/goproxy@v0.0.0-20221015165544-a0805db90819/https.go:211 +0x611
```

https://github.com/elazarl/goproxy/blob/master/https.go#L249

g0ku704:
This fix is to prevent Goproxy server crashes if the request cannot be parsed in line https://github.com/elazarl/goproxy/blob/master/https.go#L234 then logged in line https://github.com/elazarl/goproxy/blob/master/https.go#L249 when proxy is in MITM mode.

Fixes #502

### DIFF
--- a/https.go
+++ b/https.go
@@ -246,7 +246,11 @@ func (proxy *ProxyHttpServer) handleHttps(w http.ResponseWriter, r *http.Request
 						return
 					}
 					if err != nil {
-						ctx.Warnf("Illegal URL %s", "https://"+r.Host+req.URL.Path)
+						if req != nil {
+							ctx.Warnf("Illegal URL %s", "https://"+r.Host+req.URL.Path)
+						}else {
+							ctx.Warnf("Illegal URL %s", "https://"+r.Host)
+						}
 						return
 					}
 					removeProxyHeaders(ctx, req)


https.go
```diff
@@ -246,7 +246,11 @@ func (proxy *ProxyHttpServer) handleHttps(w http.ResponseWriter, r *http.Request
 						return
 					}
 					if err != nil {
-						ctx.Warnf("Illegal URL %s", "https://"+r.Host+req.URL.Path)
+						if req != nil {
+							ctx.Warnf("Illegal URL %s", "https://"+r.Host+req.URL.Path)
+						}else {
```
mpatlasov, line 251: minor nit: a space between `}` and `else` is missed

g0ku704: Fixed, thank you :+1:

mpatlasov (approved):

https.go
```diff
@@ -246,7 +246,11 @@ func (proxy *ProxyHttpServer) handleHttps(w http.ResponseWriter, r *http.Request
 						return
 					}
 					if err != nil {
-						ctx.Warnf("Illegal URL %s", "https://"+r.Host+req.URL.Path)
+						if req != nil {
```
neild, line 249: Drive by review: Should this be `if req.URL != nil`? If `req` is `nil`, then I'd expect `isWebSocketRequest` above to panic.

g0ku704: Yes, good catch. I've changed to `if req.URL != nil` instead.

kostyay:
Any plans to merge this?

g0ku704:
> Any plans to merge this?

We need a maintainer approval.

PR #333, opened 2019-04-01
Close non-persistent HTTPS connections

hmarr:
For instance, connections with the "Connection: close" header. Currently, if the client uses this header to signal a non-persistent connection and waits for the connection to be closed, it will hang indefinitely. This change closes these connections after processing the first HTTP request.

PR #514, opened 2023-10-17
send Connection: close header to client only when needed

jakecoffman:
This fixes the inability to reuse connections by only setting the Connection: close header when the client or remote server requests it.

This also pulls in changes by @hmarr in #333 to explicitly terminate the connection instead of relying on an EOF from the client.

### DIFF
--- a/https.go
+++ b/https.go
@@ -234,6 +234,9 @@ func (proxy *ProxyHttpServer) handleHttps(w http.ResponseWriter, r *http.Request
 					req.URL, err = url.Parse("https://" + r.Host + req.URL.String())
 				}
 
+				// Take the original value before filtering the request
+				closeConn := req.Close
+
 				// Bug fix which goproxy fails to provide request
 				// information URL in the context when does HTTPS MITM
 				ctx.Req = req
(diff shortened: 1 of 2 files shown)


jeffwidman:
@elazarl Sorry to bother, but I noticed this one's been floating for several months... I know the :dependabot: team currently uses a fork of Goproxy incorporating the change in #333 (which this PR includes) internally so it's already seen some prod traffic with no issues.

Any chance you could take a look? I'm sure they'd like to switch back to using upstream instead of a fork.

jakecoffman:
We actually found this change might be problematic and we're testing without it to compare. So hold off on merging for now. Thanks @jeffwidman!

jeffwidman:
Oh interesting. Since this PR includes two changes, are you observing the problem with #333 or with the other change in this PR?

elazarl:
Let me know, and I'll merge whatever you deem helpful.

Thanks!

jakecoffman:
The problem is the change in #333. We observed `write: connection reset by peer` when doing Bundler updates which went away when we removed those lines in #333. 

Now that we're running without that change we're seeing less timeouts. 

I think there are some changes in this PR that we might still want, like not setting the Close header after each request, but I'd like to run it in Dependabot production for a bit to ensure it's good first.

You can close #333!

PR #523, opened 2024-01-16
fix the problem that HTTP steam request cannot be flushed

taco-wang:


elazarl:
@taco-wang good point.
Can we parse it instead?
Shouldn't we at least use `HasPrefix`?

taco-wang:
sure, changed

PR #524, opened 2024-01-22
fix ipv6 too many colons

yaoqiangpersonal:
ipv6 address, r.URL.HOST will give an error


ErikPelli:
https://github.com/elazarl/goproxy/issues/562

PR #532, opened 2024-04-05
support for mitm ca chain

RomanManz:
This allows to return the full ca chain instead of returning only the leaf certificate.
The change is backward compatible.
Was there a security concern that I am not aware of as the reason for the initial implementation?

### DIFF
--- a/https.go
+++ b/https.go
@@ -453,6 +453,11 @@ func (proxy *ProxyHttpServer) NewConnectDialToProxyWithHandler(https_proxy strin
 }
 
 func TLSConfigFromCA(ca *tls.Certificate) func(host string, ctx *ProxyCtx) (*tls.Config, error) {
+	ca.Certificate = [][]byte{ca.Certificate[0]}
+	return TLSConfigFromCANative(ca)
+}
+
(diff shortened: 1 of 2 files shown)


https.go
```diff
@@ -453,6 +453,11 @@ func (proxy *ProxyHttpServer) NewConnectDialToProxyWithHandler(https_proxy strin
 }
 
 func TLSConfigFromCA(ca *tls.Certificate) func(host string, ctx *ProxyCtx) (*tls.Config, error) {
+	ca.Certificate = [][]byte{ca.Certificate[0]}
+	return TLSConfigFromCANative(ca)
+}
+
+func TLSConfigFromCANative(ca *tls.Certificate) func(host string, ctx *ProxyCtx) (*tls.Config, error) {
```
ErikPelli, line 460: Can't we have just a single TLSConfigFromCA function that returns the whole chain? This shouldn't be a breaking change

ErikPelli:
Thanks!
I integrated this PR into https://github.com/elazarl/goproxy/pull/574 and added you as a coauthor.

PR #536, opened 2024-05-23
Add support for KeepHeader option during HTTPS connections

aplulu:


elazarl:
Hi, 
Can you give context to the change?

Why do you need goproxy as a reverse proxy?

aplulu:
Hi,
I apologize for any confusion caused by accidentally opening a Pull Request against the upstream repository instead of my forked repository. That was an unintended mistake on my part.

To provide some context, the purpose of using goproxy as a reverse proxy in this case is to facilitate the integration of Autify, an E2E testing service, with an environment protected by Google Cloud's Identity-Aware Proxy (IAP).
The goal is to inject authentication information into the requests using the proxy created with goproxy, enabling the E2E tests to be conducted in the IAP-protected environment.
In order to achieve this, it is necessary to send the `Proxy-Authorization` header to the Google Cloud Load Balancer. Additionally, I wanted to ensure that the header is not removed even in the case of connections over HTTPS.

https://help.autify.com/docs/configuration-when-using-a-proxy-server
https://cloud.google.com/iap/docs/authentication-howto?hl=en#authenticating_from_proxy-authorization_header

PR #539, opened 2024-06-13
Add basic passthrough HTTP/2 support.

msuozzo:
Add bare-bones HTTP2 "support" that allows goproxy to permit HTTP2 traffic through but doesn't apply proxy rules.

 This also adds a dependency on the standard lib http2 package which, provided the current miniscule dep set, is worth giving some consideration.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,3 +1,10 @@
 module github.com/elazarl/goproxy
 
-require github.com/elazarl/goproxy/ext v0.0.0-20190711103511-473e67f1d7d2
+go 1.23
+
+require (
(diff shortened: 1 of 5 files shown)


elazarl:
Thanks, looks like a good start, when done we'll release another version.

roelarents:
stupid question perhaps, but does it really need go 1.23?

elazarl:
Good point. Maybe we should revert

msuozzo:
Oh shoot not sure how that change got in there. I can regenerate the PR.

msuozzo:
Alright if you'd like to fix forward, i mailed #543

Issue #403, opened 2020-09-21
Allow non-cannonical headers


physics-sec:
Hi!
When you copy the headers, using `copyHeaders` [here](https://github.com/elazarl/goproxy/blob/master/proxy.go#L38-L49), you use the `Add` method, which is defined [here](https://golang.org/src/net/http/header.go) (line 27).  

The "issue" is that `Add` canonicalizes the the headers as explained [here](https://golang.org/pkg/net/textproto/#CanonicalMIMEHeaderKey).  
This is fine for the most part, but it would be really cool to allow non-cannonical headers by passing some flag to the proxy initialization.  

In order to do that, instead of using `Add` or `Set`, assign to the map directly. As stated [here](https://golang.org/src/net/http/header.go) (line 35).  

I hope you would consider it.  
Thank you!

alfonso-presa:
Hi. I would also like to see this happening. We have some third-party client libraries that are not tolerant to case change in headers (I know that is wrong, as per the HTTP specification, they should not rely in case). Would you consider a PR for this?

Thanks!

elazarl:
Sure, would be happy for a PR

alfonso-presa:
Hi @elazarl,

trying to change this, the issue is not in the Add inside goproxy but on the httpclient itself that is making the header names cannonical. So that has to be fixed before changing goproxy behaviour.... and it doesn't seem to be happening in the sort term: https://github.com/golang/go/issues/37834

PR #553, opened 2024-11-19
Avoid Header Canonicalization in copyHeaders by Using Direct Assignment

alessiodallapiazza:
This PR updates the `copyHeaders` function to avoid the canonicalization of HTTP header keys caused by the `Add` method. Instead of using `Add`, the function now directly assigns header values from `src` to `dst`, preserving their original casing. 

Let me know if further adjustments are needed!

Closes: https://github.com/elazarl/goproxy/issues/403

### DIFF
--- a/proxy.go
+++ b/proxy.go
@@ -44,9 +44,8 @@ func copyHeaders(dst, src http.Header, keepDestHeaders bool) {
 		}
 	}
 	for k, vs := range src {
-		for _, v := range vs {
-			dst.Add(k, v)
-		}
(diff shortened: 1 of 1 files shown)


ErikPelli:
This change looks good to me for the responses, it doesn't break anything and can be easily approved.
Previously, you mentioned that there is a similar canonization issue with the requests made to the proxy, is there a solution for it that could be integrated into this project or to implement it we need a "dirty" solution as a workaround?

PR #555, opened 2024-11-28
Add support for ED25519 CA cert key

ErikPelli:
Currently `TLSConfigFromCA()` returns an error if the specified custom CA certificate uses ED25519 as the private key.
With this pull request, I add the support for it.
Compared to rsa and ecdsa, its usage among the standard library doesn't use a pointer and it has been made clear in the x509 documentation: https://pkg.go.dev/crypto/x509#MarshalPKCS8PrivateKey (it explicitly says not a pointer)

### DIFF
--- a/counterecryptor.go
+++ b/counterecryptor.go
@@ -4,6 +4,7 @@ import (
 	"crypto/aes"
 	"crypto/cipher"
 	"crypto/ecdsa"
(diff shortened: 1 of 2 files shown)


elazarl (approved):

elazarl:
Thanks!

PR #563, opened 2024-12-15
[Fix]: Adding address inside the  square brackets

### LINKED ISSUE #562: Fix ipv6 parsing
https://github.com/elazarl/goproxy/issues/336
https://github.com/elazarl/goproxy/pull/524

Currently there are some cases of reported errors that says "too many colons" when there is an IPv6 address.
In the first issue there is this value example:
```
2606:4700:30::681f:1003:443
```

For ipv6, address must be inside square brackets, and in order to be parsed it must be:
```
[2606:4700:30::681f:1003]:443
```

Check where is the problem and fix it accordingly, in order to correct use the square brackets.

nakul-py:
- ***This PR fixes the issue where IPv6 addresses with ports were not being correctly parsed due to an incorrect handling of colons.***
 - ***The fix ensures that the address is wrapped in square brackets and the port is properly separated.***

### DIFF
--- a/https.go
+++ b/https.go
@@ -51,24 +51,19 @@ func stripPort(s string) string {
 	var ix int
 	if strings.Contains(s, "[") && strings.Contains(s, "]") {
 		//ipv6 : for example : [2606:4700:4700::1111]:443
-
-		//strip '[' and ']'
-		s = strings.ReplaceAll(s, "[", "")
-		s = strings.ReplaceAll(s, "]", "")
-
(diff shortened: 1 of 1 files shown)


ErikPelli:
This is not the function that needs to be changed to fix the issue.
The problem is related to the request, in the reported issues linked in #562.
Here you are changing stripPort(), which is used only to extract the hostname for the certificate generation / cache retrieval.
If you find the correct solution, you can open another pull request. Closing this one.

Issue #560, opened 2024-12-14
Add instructions about local certificate usage / certification authority trust


ErikPelli:
There are a lot of issues (like https://github.com/elazarl/goproxy/issues/142 and https://github.com/elazarl/goproxy/issues/529) where the users complain that the proxy is not accepted by their browser because their certificate authority is not among the trusted ones.
We should add some sections in the documentation where we explain how to do that.
Related to https://github.com/elazarl/goproxy/issues/222.

Basically, some instructions about how to add the existing goproxy certificate to the trusted ones by the browser or to generate a new valid one and trust it (and how to use it in goproxy, citing the already existing custom CA example).

You can add all the instructions to the README of the custom CA example.

ErikPelli:
https://github.com/elazarl/goproxy/issues/417

ErikPelli:
This issue is open for anyone

PR #573, opened 2024-12-18
Add example for SOCKS proxy chaining

yuisofull:
I add  an example demonstrating how to configure goproxy to forward both HTTP and HTTPS traffic through a SOCKS5 proxy.  Since the SOCKS5 dialer doesn't handle DNS resolution, I manually resolve the target addresses before passing them to the SOCKS5 proxy.


ErikPelli:
Instead of using net/proxy, can you directly use the standard library (https://pkg.go.dev/net/http#Transport), where they states that support socks5 as a proxy schema?
It's enough to change the transport with one that has this proxy function that returns a socks URL

ErikPelli:
Maybe you can make a variant of cascadeproxy where the end proxy is a socks server, it's up to you

ErikPelli:
Oh, last thing, I've rewritten the cascadeproxy in the rewrite-examples branch. [rewrite-examples](https://github.com/elazarl/goproxy/tree/rewrite-examples)

If you decide to base the code on this example, use the code from this branch and not from the master, please

yuisofull:
Thanks a lot for your help, I'll try to figure it out

yuisofull:
Thanks alot for your help. I'm having issue on creating ConnectDial without the socks client

ErikPelli:
What's the issue?

yuisofull:
Im doing sth like:
```go
proxyServer.Tr.Proxy = func(r *http.Request) (*url.URL, error) {
		Url := &url.URL{
			Scheme: "socks5",
			Host:   socksAddr,
		}
		if auth.User != "" {
			Url.User = url.UserPassword(auth.User, auth.Password)
		}
		return Url, nil
	}
```
Without the ConnectDial, it will only redirect the http request to socks5. And in the case that instead of using net/proxy, Ill try to do the same thing with func (proxy *ProxyHttpServer) NewConnectDialToProxyWithHandler, pass in the socks5 proxy address, and then implement the raw handshake with socks and connect to it

ErikPelli:
Yes, ConnectDial is required, this morning I updated the cascadeproxy example to include it, I confirm that it's required

yuisofull:
I have rewritten the code and implemented the ConnectDial

examples/goproxy-socks/socksproxy.go
```diff
@@ -0,0 +1,73 @@
+// This example demonstrates how to configure goproxy to act as an HTTP/HTTPS proxy
+// that forwards all traffic through a SOCKS5 proxy.
+// The goproxy server acts as an aggregator, handling incoming HTTP and HTTPS requests
+// and routing them via the SOCKS5 proxy.
+// Example usage:
+// socks proxy with no auth:
+// 		go run socksproxy.go -v -addr ":8080" -socks "localhost:1080"
+// socks with auth:
+// 		go run socksproxy.go -v -addr ":8080" -socks "localhost:1080" -user "bob" -pass "123"
```
ErikPelli, line 9: Can you move all this part into a README file like all the other examples, maybe adding some more explaination that this is based on cascadeproxy, etc.?

examples/goproxy-socks/socksproxy.go
```diff
@@ -0,0 +1,73 @@
+// This example demonstrates how to configure goproxy to act as an HTTP/HTTPS proxy
+// that forwards all traffic through a SOCKS5 proxy.
+// The goproxy server acts as an aggregator, handling incoming HTTP and HTTPS requests
+// and routing them via the SOCKS5 proxy.
+// Example usage:
+// socks proxy with no auth:
+// 		go run socksproxy.go -v -addr ":8080" -socks "localhost:1080"
```
ErikPelli, line 7: Instead of go run, use go build and then `./insert-here-compiled-name`

ErikPelli:
Can you call the exampe `cascadeproxy-socks` instead of the current name?

examples/goproxy-socks/socksproxy.go
```diff
@@ -0,0 +1,73 @@
...
+
+	proxyServer.OnRequest().Do(goproxy.FuncReqHandler(func(req *http.Request, ctx *goproxy.ProxyCtx) (*http.Request, *http.Response) {
+		resolvedAddr, _ := net.ResolveIPAddr("ip", req.URL.Host)
+		req.URL.Host = resolvedAddr.String() + ":" + req.URL.Port()
+		return req, nil
+	}))
+	proxyServer.Tr.Proxy = createSocksProxy(*socksAddr, auth)
+	proxyServer.Tr.Dial = func(network, addr string) (net.Conn, error) {
+		return net.Dial(network, addr)
+	}
```
ErikPelli, line 58: I don't think this is needed

examples/goproxy-socks/socksproxy.go
```diff
@@ -0,0 +1,73 @@
...
+	socksConnectDial, err := NewSocks5ConnectDialedToProxy(proxyServer, *socksAddr, &auth, func(req *http.Request) {
+		if strings.ContainsRune(req.URL.Host, ':') {
+			req.URL.Host = strings.Split(req.URL.Host, ":")[0]
+		}
+		resolvedAddr, _ := net.ResolveIPAddr("ip", req.URL.Host)
+		req.URL.Host = resolvedAddr.String() + ":" + req.URL.Port()
+	})
```
ErikPelli, line 65: Remove this parse and resolve part, when go dialer receives an hostname as the host address, it automatically resolves it, no need to manually do it

examples/goproxy-socks/socksclient.go
```diff
@@ -0,0 +1,253 @@
...
+	}
+	if 1 > portnum || portnum > 0xffff {
+		return "", 0, errors.New("port number out of range " + port)
+	}
+	ip, err := net.ResolveIPAddr("ip", host)
+	if err != nil {
+		return "", 0, err
+	}
+	return ip.String(), portnum, nil
+}
```
ErikPelli, line 253: Instead of copying the whole file, isn't there something exposed in that packages (or other packages of standard library) that just allows the opening of a dialed connection through a target SOCKS server?

examples/goproxy-socks/socksproxy.go
ErikPelli: Not sure about this, is it specified as something already present in the standard library? I've just written this in the review, but if its not suported, just let me know

yuisofull:
I dont know why but in my case if I dont resolve the ip address myself it would have something like this:
`socks connect tcp 127.0.0.1:1080->www.google.com:443: unknown error host unreachable`

yuisofull:
with the standard library I can do something like(work for both http and https):
```go
package main

import (
	"crypto/tls"
	"flag"
	"fmt"
	"io/ioutil"
	"log"
	"net"
	"net/http"
	"net/url"
	"strings"
	"time"
)

type SocksAuth struct {
	User, Password string
}

func createSocksProxy(socksAddr string, auth SocksAuth) func(r *http.Request) (*url.URL, error) {
	return func(r *http.Request) (*url.URL, error) {
		host := r.URL.Host
		if !strings.ContainsRune(host, ':') {
			host += ":" + r.URL.Port()
		}
		resolvedAddr, _ := resolveAddress(host)
		r.URL.Host = resolvedAddr
		Url := &url.URL{
			Scheme: "socks5",
			Host:   socksAddr,
		}
		if auth.User != "" {
			Url.User = url.UserPassword(auth.User, auth.Password)
		}
		return Url, nil
	}
}
func main() {
	target := flag.String("target", "https://www.google.com", "URL to get")
	proxyAddr := flag.String("proxy", "localhost:1080", "SOCKS5 proxy address to use")
	username := flag.String("user", "", "username for SOCKS5 proxy")
	password := flag.String("pass", "", "password for SOCKS5 proxy")
	flag.Parse()

	auth := SocksAuth{
		User:     *username,
		Password: *password,
	}
	client := &http.Client{
		Transport: &http.Transport{
			Proxy: createSocksProxy(*proxyAddr, auth),
			TLSClientConfig: &tls.Config{
				InsecureSkipVerify: true,
			},
		},

		Timeout: 30 * time.Second,
	}
	resolvedAddr, err := resolveAddress(*target)
	if err != nil {
		log.Fatal(err)
	}
	req, err := http.NewRequest("GET", *target, nil)
	if err != nil {
		log.Fatalf("Failed to create request: %v", err)
	}
	req.URL.Host = resolvedAddr
	r, err := client.Do(req)
	if err != nil {
		log.Fatal(err)
	}
	defer r.Body.Close()
	body, err := ioutil.ReadAll(r.Body)
	if err != nil {
		log.Fatal(err)
	}
	fmt.Println(string(body))
}

func resolveAddress(target string) (string, error) {
	u, err := url.Parse(target)
	if err != nil {
		return "", fmt.Errorf("invalid URL: %w", err)
	}
	host, port := u.Hostname(), u.Port()

	ipAddr, err := net.ResolveIPAddr("ip", host)
	if err != nil {
		return "", fmt.Errorf("failed to resolve hostname: %w", err)
	}

	return net.JoinHostPort(ipAddr.String(), port), nil
}
```
But I havent figured out how to implement ConnectDial with this

ErikPelli:
> with the standard library I can do something like(work for both http and https):
> 
> ...
> But I havent figured out how to implement ConnectDial with this

We can try to use a different approach:
1. Intercept the HTTPS request using MITM and avoid to call the destination
2. When we handle the MITM'd request, we obtain an already parsed req
3. With this req, instead of let the proxy handle it, we can create a custom HTTP client/transport with the address of the SOCKS proxy specified as proxy
4. Instead of the default client, we use this custom http client to make the request

What do you think?

yuisofull:
Hmm, Ill try this approach

ErikPelli:
If you want, you can also launch a test local SOCKS5 proxy in your example, using this library: https://github.com/things-go/go-socks5

yuisofull:
Im having a local socks5 proxy using dante on linux

ErikPelli:
> Im having a local socks5 proxy using dante on linux

Yes, but the example should be used by anyone that wants to.
Embedding a socks server inside it would make it easier for the users that want to test it.

ErikPelli:
> HijackConnect
Don't. It's not good for this use.

yuisofull:
I have removed the socks5 client

examples/cascadeproxy-socks/socksproxy.go
```diff
@@ -0,0 +1,99 @@
...
+		req.URL.Host = resolvedAddr.String() + ":" + req.URL.Port()
+		return req, nil
+	}))
+
+	proxyServer.OnRequest().HandleConnect(goproxy.FuncHttpsHandler(func(host string, ctx *goproxy.ProxyCtx) (*goproxy.ConnectAction, string) {
+		req := ctx.Req
+		req.RequestURI = ""
+		client := &http.Client{
+			Transport: &http.Transport{
+				Proxy: createSocksProxy(*socksAddr, auth),
```
ErikPelli, line 56: All this logic should be inside the Do / DoFunc function!
The connect handler must be only `goproxy.AlwaysMitm`.
This way the proxy server will take over the connection.

Then, in the Do function, you return the resulting *http.Response as the second parameter, and the server will return it directly to the client.

yuisofull: Im having this error:

```go
	proxyServer.OnRequest().HandleConnect(goproxy.FuncHttpsHandler(func(host string, ctx *goproxy.ProxyCtx) (*goproxy.ConnectAction, string) {
		return &goproxy.ConnectAction{
			Action: goproxy.ConnectMitm,
		}, host
	}))
```

```
2024/12/20 20:25:20 [001] INFO: Running 1 CONNECT handlers
2024/12/20 20:25:20 [001] INFO: on 0th handler: &{2 <nil> <nil>} www.google.com:443
2024/12/20 20:25:20 [001] INFO: Assuming CONNECT is TLS, mitm proxying it
2024/12/20 20:25:20 [001] WARN: Cannot handshake client www.google.com:443 tls: no certificates configured
```

Its immediately  stop when its do the handshake

```go
	case ConnectMitm:
		proxyClient.Write([]byte("HTTP/1.0 200 OK\r\n\r\n"))
		ctx.Logf("Assuming CONNECT is TLS, mitm proxying it")
		// this goes in a separate goroutine, so that the net/http server won't think we're
		// still handling the request even after hijacking the connection. Those HTTP CONNECT
		// request can take forever, and the server will be stuck when "closed".
		// TODO: Allow Server.Close() mechanism to shut down this connection as nicely as possible
		tlsConfig := defaultTLSConfig
		if todo.TLSConfig != nil {
			var err error
			tlsConfig, err = todo.TLSConfig(host, ctx)
			if err != nil {
				httpError(proxyClient, ctx, err)
				return
			}
		}
		go func() {
			//TODO: cache connections to the remote website
			rawClientTls := tls.Server(proxyClient, tlsConfig)
			defer rawClientTls.Close()
			if err := rawClientTls.Handshake(); err != nil {
				ctx.Warnf("Cannot handshake client %v %v", r.Host, err)
				return
			}
```

ErikPelli: Instead of
```
&goproxy.ConnectAction{
			Action: goproxy.ConnectMitm,
		}
```

Please use
```
goproxy.AlwaysMitm
```

yuisofull: I solved it with goproxy.AlwaysMitm but I encountered another issue:

```
curl: (60) SSL certificate problem: self-signed certificate in certificate chain
More details here: https://curl.se/docs/sslcerts.html

curl failed to verify the legitimacy of the server and therefore could not
establish a secure connection to it. To learn more about this situation and
how to fix it, please visit the web page mentioned above.
```

ErikPelli: The alternative to -k is the guide to trust the certificates, but no one wrote it yet :)
https://github.com/elazarl/goproxy/issues/560

yuisofull:
> All this logic should be inside the Do / DoFunc function! The connect handler must be only `goproxy.AlwaysMitm`. This way the proxy server will take over the connection.

Done!

examples/cascadeproxy-socks/socksproxy.go
```diff
@@ -0,0 +1,98 @@
...
+
+import (
+	"crypto/tls"
+	"flag"
+	"github.com/elazarl/goproxy"
+	"log"
+	"net"
+	"net/http"
+	"net/url"
+	"strings"
```
ErikPelli, line 11:
```suggestion
	"crypto/tls"
	"flag"
	"log"
	"net"
	"net/http"
	"net/url"
	"strings"

	"github.com/elazarl/goproxy"
```

examples/cascadeproxy-socks/socksproxy.go
```diff
@@ -0,0 +1,98 @@
...
+		}
+
+		ip, err := net.ResolveIPAddr("ip", host)
+		if err != nil {
+			ctx.Logf("Failed to resolve host: " + err.Error())
+			return nil, nil
+		}
+		host = net.JoinHostPort(ip.String(), port)
+		req.URL.Host = host
+		req.Host = host
```
ErikPelli, line 82: No need for this scheme set and ip resolving part, just define client and then to client.Do

ErikPelli (changes requested):

examples/cascadeproxy-socks/socksproxy.go
yuisofull: aye I forgot it

yuisofull:
> No need for this scheme set and ip resolving part, just define client and then to client.Do

Done

examples/cascadeproxy-socks/socksproxy.go
```diff
@@ -0,0 +1,69 @@
...
+		if err != nil {
+			ctx.Logf("Failed to forward request: " + err.Error())
+			return nil, nil
+		}
+		ctx.Logf("Succesfully forwarded request to socks proxy")
+		return req, resp
+	})
+
+	proxyServer.OnRequest().HandleConnect(goproxy.AlwaysMitm)
+	proxyServer.Tr.Proxy = createSocksProxy(*socksAddr, auth)
```
ErikPelli, line 65: No need to set the proxy here, since you use the custom client

examples/cascadeproxy-socks/socksproxy.go
```diff
@@ -0,0 +1,69 @@
...
+		client := &http.Client{
+			Transport: &http.Transport{
+				Proxy: createSocksProxy(*socksAddr, auth),
+				TLSClientConfig: &tls.Config{
+					InsecureSkipVerify: true,
+				},
+			},
+		}
+
+		req.RequestURI = ""
```
ErikPelli, line 54: Any reason to set an empty RequestURI?

examples/cascadeproxy-socks/README.md
```diff
@@ -0,0 +1,29 @@
...
+the requests to another socks proxy server. This example is written base on `cascadeproxy` example.
+
+Diagram:
+```
+client --> goproxy --> socks5 proxy --> internet
+```
+
+This example starts a HTTP/HTTPS proxy using goproxy that listens on port `8080`, and forward the requests to the socks5 proxy on `socks5://localhost:1080`.
+
+
```
ErikPelli, line 13: Can you mention the fact that we do MITM to use the SOCKS proxy and that the proxy forwarding is impemented by the Go standard library?

examples/cascadeproxy-socks/socksproxy.go
yuisofull: Yes, otherwise it will result in:

```
Get "http://www.example.com/": http: Request.RequestURI can't be set in client requests
```

yuisofull:
> Can you mention the fact that we do MITM to use the SOCKS proxy and that the proxy forwarding is impemented by the Go standard library?

Ive added

yuisofull:
> Ok, you can add a comment about it where you cite this sof answer: https://stackoverflow.com/questions/19595860/http-request-requesturi-field-when-making-request-in-go

Ive added

examples/cascadeproxy-socks/socks5proxyserver/go.mod
```diff
@@ -0,0 +1,5 @@
+module socks5proxyserver
+
+go 1.23.4
```
ErikPelli, line 3: Use go 1.20 as the version

examples/cascadeproxy-socks/socksproxy.go
ErikPelli: What about this?

yuisofull: Ive removed it

examples/cascadeproxy-socks/README.md
```diff
@@ -0,0 +1,29 @@
...
+`CascadeSocksProxy` is an example that shows an aggregator server that forwards
+the requests to another socks proxy server. This example is written base on `cascadeproxy` example.
+
+Diagram:
+```
+client --> goproxy --> socks5 proxy --> internet
+```
+
+This example starts a HTTP/HTTPS proxy using goproxy that listens on port `8080`, and forward the requests to the socks5 proxy on `socks5://localhost:1080`.
+Its use MITM to forward the HTTP/HTTPS requests to socks5 proxy with Go standard library `net/http`.
```
ErikPelli, line 12:
```suggestion
It uses MITM (Man in the Middle) proxy mode to retriece and parse the request, and then forwards it to the destination using  the socks5 proxy client implemented in the standard Go `net/http` library. 
```

ErikPelli:
Thanks! Now it looks good to me. Sorry If I have bothered you with all the requested changes!

PR #578, opened 2024-12-19
Rewrite readme

ErikPelli:
Rewrite project readme to include new updates and make it more appropriate.
This PR also updates x/net dependency.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,7 @@ proxy is `localhost:8080`, which is the default one in our example.
 - Use a `custom http.Transport` to perform requests to the target server
 - You can specify a `MITM certificates cache`, to reuse them later for other requests to the same host, thus saving CPU. Not enabled by default, but you should use it in production!
 - Redirect normal HTTP traffic to a `custom handler`, when the target is a `relative path` (e.g. `/ping`)
+- You can choose the logger to use, by implementing the `Logger` interface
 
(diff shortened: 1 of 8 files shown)


ErikPelli:
@elazarl what do you think about this new readme? Can it be merged?
Here there is a link with the formatted preview so you can check it better: https://github.com/elazarl/goproxy/blob/5e122f936680ff0d35271525e2791b3ce47b095d/README.md

elazarl:
@ErikPelli truly amazing.  IMHO _italics_ emphasizes better than `code`.

Let's merge!

ErikPelli:
Do you prefer italic style? I could change it if you want

elazarl:
> Do you prefer italic style? I could change it if you want

Let me explain

> The proxy itself is simply a net/http handler,

Here it's great to have `net/http` as code, as this is a piece of code.

> The target of this project is to offer an optimized proxy server,

Here _optimized_ is more appropriate, as it is an English word, and not a piece of code that could appear in the source.

Does that make sense?

elazarl:
NM, that's a cool style, and let's keep it as it is for the time being. I like it when your fingerprints are felt over the project.

I'll take the oppurtunity to thank you for the tremndous work and effort you invest in the project. It is felt and is hearwarming.

PR #610, opened 2024-12-31
ext/har: add HAR logger extension

### LINKED ISSUE #609: HAR dumper
Port the HAR logger feature from [abourget/goproxy](https://github.com/elazarl/goproxy/compare/master...abourget:goproxy:master) to the `ext/` package of this project.
This feature must be part of the external package and not do a mess with the core project.

CameronBadman:
ext/har: add HAR logger extension

Port HAR (HTTP Archive) logging functionality from abourget/goproxy
to the ext/ package. Features include:
- Request and response capture
- Content body logging with toggle
- HAR 1.2 spec compliance
- Thread-safe operation
- Comprehensive test coverage

This is a self-contained extension that doesn't modify core functionality.
Credit to abourget/goproxy for the original implementation.

Closes #609


ext/har/types.go
```diff
@@ -0,0 +1,365 @@
...
+	if resp.ContentLength == 0 {
+		log.Println("Empty content")
+		return
+	}
+
+	body, _ := ioutil.ReadAll(resp.Body)
+	harContent.Text = string(body)
+	harContent.Size = len(body)
+	return
+}
```
ErikPelli, line 309: This function is not used, should we remove it?

ext/har/types.go
```diff
@@ -0,0 +1,365 @@
...
+	}
+	return harPostData
+}
+
+func parseStringArrMap(stringArrMap map[string][]string) []NameValuePair {
+	index := 0
+	harQueryString := make([]NameValuePair, len(stringArrMap))
+	for k, v := range stringArrMap {
+		escapedKey, _ := url.QueryUnescape(k)
+		escapedValues, _ := url.QueryUnescape(strings.Join(v, ","))
```
ErikPelli, line 205: QueryUnescape will return an empty string if there is an error, in this case we should leave the original value if we are unable to unescape it, I think it's better than an empty string

ext/har/types.go
CameronBadman: I thought it should be kept in for future development, but sure I'll remove it

ext/har/types.go
```diff
@@ -0,0 +1,365 @@
...
+func calcHeaderSize(header http.Header) int64 {
+	headerSize := 0
+	for headerName, headerValues := range header {
+		headerSize += len(headerName) + 2
+		for _, v := range headerValues {
+			headerSize += len(v)
+		}
+	}
+	return int64(headerSize)
+}
```
ErikPelli, line 167: This calculation is an huge approximation and it's not reliable.
We should follow the HAR specs and use `-1` as the headersSize value instead, because we don't have this information.

ext/har/types.go
```diff
@@ -0,0 +1,365 @@
...
+	}
+
+	if captureContent && (req.Method == "POST" || req.Method == "PUT") {
+		harRequest.PostData = parsePostData(req)
+	}
+
+	return &harRequest
+}
+
+func (harEntry *Entry) FillIPAddress(req *http.Request) {
```
ErikPelli, line 139: Keep this function private
```suggestion
func (harEntry *Entry) fillIPAddress(req *http.Request) {
```

ext/har/types.go
```diff
@@ -0,0 +1,365 @@
...
+	host, _, err := net.SplitHostPort(req.URL.Host)
+	if err != nil {
+		host = req.URL.Host
+	}
```
ErikPelli, line 143: Use req.URL.Hostname() instead of this

ext/har/types.go
```diff
@@ -0,0 +1,365 @@
...
+
+	if ipaddr, err := net.LookupIP(host); err == nil {
+		for _, ip := range ipaddr {
+			if ip.To4() != nil {
+				harEntry.ServerIpAddress = ip.String()
+				return
+			}
+		}
+	}
```
ErikPelli, line 155: Since this is an optional field, avoid to manually make a DNS resolution and handle only the previous case where we directly have the IP address.
If we manually resolve the IP, the result can be different from the originally resolved IP and will cause confusion.

ext/har/types.go
```diff
@@ -0,0 +1,365 @@
...
+	}()
+
+	harPostData := new(PostData)
+	contentType := req.Header["Content-Type"]
+	if contentType == nil {
+		panic("Missing content type in request")
+	}
+	harPostData.MimeType = contentType[0]
+
+	if len(req.PostForm) > 0 {
```
ErikPelli, line 183: By reading the documentation, PostForm is available only when we have previously called ParseForm(), but here it's missing.
https://pkg.go.dev/net/http#Request

ext/har/types.go
```diff
@@ -0,0 +1,365 @@
...
+
+func parsePostData(req *http.Request) *PostData {
+	defer func() {
+		if e := recover(); e != nil {
+			log.Printf("Error parsing request to %v: %v\n", req.URL, e)
+		}
+	}()
+
+	harPostData := new(PostData)
+	contentType := req.Header["Content-Type"]
```
ErikPelli, line 177: Use Header.Get instead of raw map get

ext/har/types.go
```diff
@@ -0,0 +1,365 @@
...
+	if contentType == nil {
+		panic("Missing content type in request")
+	}
+	harPostData.MimeType = contentType[0]
```
ErikPelli, line 181: Instead of this mess with the panic, just don't add the form data to the request and do nothing, by returning early

ext/har/types.go
```diff
@@ -0,0 +1,365 @@
...
+    "io/ioutil"
+    "log"
+    "net"
+    "net/http"
+    "net/url"
+    "strings"
+    "time"
+)
+
+var startingEntrySize int = 1000
```
ErikPelli, line 17: Instead of a global variable, add a
const startingEntrySize = 1000
inside `makeNewEntries`

ext/har/types.go
```diff
@@ -0,0 +1,365 @@
...
+				param := PostDataParam{
+					Name:  k,
+					Value: v,
+				}
+				harPostData.Params = append(harPostData.Params, param)
+			}
+		}
+	} else {
+		str, _ := ioutil.ReadAll(req.Body)
+		harPostData.Text = string(str)
```
ErikPelli, line 195: ioutil is deprecated, use io.ReadAll instead

ext/har/logger.go
```diff
@@ -0,0 +1,110 @@
...
+
+	return resp
+}
+
+// SetCaptureContent enables or disables request/response body capture
+func (l *Logger) SetCaptureContent(capture bool) {
+	l.mu.Lock()
+	defer l.mu.Unlock()
+	l.captureContent = capture
+}
```
ErikPelli, line 78: This functionality complicates the code, just remove the possibility to enable/disable it and keep it always running

ext/har/logger.go
```diff
@@ -0,0 +1,110 @@
...
+		return resp
+	}
+
+	// Create HAR entry
+	entry := Entry{
+		StartedDateTime: startTime,
+		Time:           time.Since(startTime).Milliseconds(),
+		Request:        ParseRequest(ctx.Req, l.captureContent),
+		Response:       ParseResponse(resp, l.captureContent),
+		Cache:          Cache{},
```
ErikPelli, line 54: Instead of setting an empty object, just skip the struct value and the zero value will remain by default

ext/har/logger.go
```diff
@@ -0,0 +1,110 @@
...
+	defer l.mu.Unlock()
+
+	file, err := os.Create(filename)
+	if err != nil {
+		return err
+	}
+	defer file.Close()
+
+	encoder := json.NewEncoder(file)
+	encoder.SetIndent("", "  ")
```
ErikPelli, line 92: No need to add json indentation, just use json.Marshal

ext/har/logger_test.go
```diff
@@ -0,0 +1,114 @@
...
+    // Verify GET request
+    if harLog.Log.Entries[0].Request.Method != "GET" {
+        t.Errorf("Expected GET request, got %s", harLog.Log.Entries[0].Request.Method)
+    }
+
+    // Verify POST request
+    if harLog.Log.Entries[1].Request.Method != "POST" {
+        t.Errorf("Expected POST request, got %s", harLog.Log.Entries[1].Request.Method)
+    }
+}
```
ErikPelli, line 114: We are using testify library for the new tests, can you refactor this test by integrating it? Like we have done in the header canonicalization

ext/har/logger.go
```diff
@@ -0,0 +1,110 @@
...
+}
+
+// GetEntries returns a copy of the current HAR entries
+func (l *Logger) GetEntries() []Entry {
+	l.mu.Lock()
+	defer l.mu.Unlock()
+	entries := make([]Entry, len(l.har.Log.Entries))
+	copy(entries, l.har.Log.Entries)
+	return entries
+}
```
ErikPelli, line 110: The HAR exporting part must be redesigned from scratch, here there are a bunch of issues.
Let me know if you want to do that, or in case I can fix this personally.

Basically, instead of this, the most common use will be the periodically send (e.g. every 5 minutes) of the HAR entries to an external logger API service or graphana or whatever.
So we have to accept a function from the user that, periodically, will consume the list of entries, pass it to this user function, and delete it from our queue.
The user will do whatever he wants with this list of entries, because he passed us its own function.

A goroutine will read from the entries channel and append to a local slice.
For the function call, there can be two options that the user can select:
- every N requests we call the function (we implement this using a counter)
- every time.Duration (e.g. 5 minutes) we call the user function

Let me know about this

ext/har/types.go
ErikPelli: Its functionality is already implemented inside another function in the same file, it can be safely removed, probably it's legacy code

ext/har/logger.go
CameronBadman: I'm actually fixing all the issues you have pointed out right now,  I can do this part tonight, but if not tomorrow. I haven't worked with go routines before, but I think I understand the general idea for the saving function. thank you.

ErikPelli: Yeah don't worry, it's holiday time, so feel free to take your time.
Your help is really appreciated here!
It's not an urgent feature.

CameronBadman: nah I should be thanking you! the feedback is appreciated. 

> Yeah don't worry, it's holiday time, so feel free to take your time. Your help is really appreciated here! It's not an urgent feature.

ext/har/logger_test.go
ErikPelli: I merged the canonicalization pull request right now, please rebase over the new master branch to implement this change

ext/har/logger.go
CameronBadman: @ErikPelli I'm going to assume that you don't want the option to disable it and I should remove the function and have it true on initialization.

ErikPelli: You should completely remove the variable and assume it's always true, yes

ext/har/logger.go
CameronBadman: @ErikPelli went ahead and updated the code to the specification you asked for, I found it actually pretty difficult, but I think it works correctly.

ext/har/logger_test.go
CameronBadman: @ErikPelli I decided updating via a pull request would be a better idea instead of doing a rebase, as I don't have any experience with it rebasing and don't want to risk issues. Hope to hear from you soon

ext/har/types.go
```diff
@@ -0,0 +1,414 @@
...
+        }
+    }
+    
+    // If no IPv4 address found, use the first IP (IPv6) in the list
+    if len(ips) > 0 {
+        entry.ServerIpAddress = ips[0].String()
+    } else {
+        // If no IPs found, fall back to the hostname
+        entry.ServerIpAddress = host
+    }
```
ErikPelli, line 216: Server IP address is an optional field, leave it empty if net.ParseIP() fails, as I explained in the previous comment (https://github.com/elazarl/goproxy/pull/610#discussion_r1900122992).
Don't call net.Resolver at all.

ext/har/types.go
```diff
@@ -0,0 +1,414 @@
...
+    mediaType, _, err := mime.ParseMediaType(contentType)
+    if err != nil {
+        log.Printf("Error parsing media type: %v", err)
+        return nil
+    }
+    
+    harPostData.MimeType = mediaType
+    
+    if err := req.ParseForm(); err != nil {
+        log.Printf("Error parsing form: %v", err)
```
ErikPelli, line 236: Instead of using "log" package, directly use the proxy logger.
You can access it by using ctx.Proxy.Logger.
Apply this change also to the other logging calls.

ext/har/types.go
```diff
@@ -0,0 +1,414 @@
...
+
+func ParseResponse(resp *http.Response, captureContent bool) *Response {
+    if resp == nil {
+        return nil
+    }
+
+    statusText := resp.Status
+    if len(resp.Status) > 4 {
+        statusText = resp.Status[4:]
+    }
```
ErikPelli, line 329: Instead of this manual stuff, use http.StatusText(resp.StatusCode), directly inside the Response{} creation in the next lines

ext/har/types.go
```diff
@@ -0,0 +1,414 @@
...
+	har.Log.Entries = append(har.Log.Entries, entry...)
+}
+
+func (har *Har) AppendPage(page ...Page) {
+	har.Log.Pages = append(har.Log.Pages, page...)
+}
+
+func makeNewEntries() []Entry {
+    const startingEntrySize int = 1000;
+	return make([]Entry, 0, startingEntrySize)
```
ErikPelli, line 57: Remove ;
Never use semicolons!

Reformat the lines

ext/har/types.go
```diff
@@ -0,0 +1,414 @@
...
+	return har
+}
+
+func (har *Har) AppendEntry(entry ...Entry) {
+	har.Log.Entries = append(har.Log.Entries, entry...)
+}
+
+func (har *Har) AppendPage(page ...Page) {
+	har.Log.Pages = append(har.Log.Pages, page...)
+}
```
ErikPelli, line 53: Why are you exposing these functions if they are never used and the Har struct members are public?

ext/har/logger.go
```diff
@@ -0,0 +1,185 @@
...
+    }
+    
+    jsonData, err := json.Marshal(har)
+    if err != nil {
+        return err
+    }
+    
+    _, err = file.Write(jsonData)
+    return err
+}
```
ErikPelli, line 168: Delete this function, it's unused

ext/har/logger.go
```diff
@@ -0,0 +1,185 @@
...
+
+// GetEntries returns a copy of the current HAR entries
+func (l *Logger) GetEntries() []Entry {
+    l.mu.Lock()
+    defer l.mu.Unlock()
+    entries := make([]Entry, len(l.entries))
+    copy(entries, l.entries)
+    return entries
+}
```
ErikPelli, line 185: Delete these two functions, user will handle entries only inside its custom handler

ext/har/logger.go
ErikPelli: @CameronBadman you're doing a 100 milliseconds polling with mutex locks, this will dangerously increase the CPU usage and it's not the correct approach for Go.
You should create a channel where the OnReponse() function can send the parsed data.
The function that you called exportLoop() will then contonuously collect from this channel (no polling at all).
If the user specified a duration, in the exportLoop() you will have a select between a ticker with this duration and the data channel read.
When you read from the data channel, you append to a local response slice.
If the user specified a length amount (> 0) and it has been reached, or If the ticker has been triggered and length is > 0, you pass the slice to the callback and set the value of the local variable to nil.
You don't need mutexes at all with this approach, nor polling.

ext/har/logger.go
```diff
@@ -0,0 +1,185 @@
...
+type ExportFunc func([]Entry)
+
+// Logger implements a HAR logging extension for goproxy
+type Logger struct {
+    mu             sync.Mutex
+    entries        []Entry
+    captureContent bool
+    exportFunc     ExportFunc
+    exportInterval time.Duration
+    exportCount    int
```
ErikPelli, line 22:
```suggestion
    exportThreshold    int
```

ext/har/logger.go
```diff
@@ -0,0 +1,185 @@
...
+    currentCount   int
+    lastExport     time.Time
```
ErikPelli, line 24: Remove these variables, they are useless

ext/har/logger_test.go
```diff
@@ -0,0 +1,268 @@
...
+            time.Sleep(200 * time.Millisecond)
+
+            // Verify HAR entry
+            entries := logger.GetEntries()
+            require.Len(t, entries, 1, "Should have one log entry")
+            entry := entries[0]
+            assert.Equal(t, tc.expectedMethod, entry.Request.Method, "Request method should match")
+
+            // Verify exported entries
+            assert.Len(t, exportedEntries, 0, "Should not have exported entries yet")
```
ErikPelli, line 110: Big NO, don't call GetEntries, this will lead to flaky tests.
Instead, you can add the test entries directly inside the callback handler, and it will be deterministic.

ext/har/logger_test.go
```diff
@@ -0,0 +1,268 @@
...
+    // Send 3 requests
+    for i := 0; i < 3; i++ {
+        resp, err := client.Get(background.URL)
+        require.NoError(t, err, "Should send request")
+        resp.Body.Close()
+        time.Sleep(200 * time.Millisecond)
+    }
+
+    // Wait for export interval
+    time.Sleep(600 * time.Millisecond)
```
ErikPelli, line 234: Remove time.Sleep from all your tests and use testify directly inside the callback handler to have a determinisitc behavior (maybe using request counter instead of time based calllback call).
And maybe use a sync.WaitGroup to wait for the finish of the callback, to make sure it's executed.

ErikPelli:
@CameronBadman any update about the last comment I left?

ext/har/types.go
CameronBadman: @ErikPelli thought I should confirm:
1. to confirm, I would  use `*goproxy.ProxyCtx` as an argument in the logging functions to use ctx?
2. also, could it be a good idea to expose the internal `printf` method in ctx?
```go
func (ctx *ProxyCtx) printf(msg string, argv ...any) {
    ctx.Proxy.Logger.Printf("[%03d] "+msg+"\n", append([]any{ctx.Session & 0xFFFF}, argv...)...)
}

Comment (ErikPelli) [reply ext/har/types.go:236]: printf() must remain for core internal use only.
You can use the Printf() method of the logger without the additional data.
In the ext/har package, logging is only to report errors and is very rare, we don't need to track additional information like the session.

**Comment by CameronBadman** [comment]
@ErikPelli that should be to the specification you wanted now, it was a good learning experience. Although I do have 1 question, with the data channel (dataCh), to my understanding I have to set a limited space amount for it, I wasn't sure what to set this at, any recommendations?

File: ext/har/logger.go [hunk 28]
```diff
@@ -0,0 +1,157 @@
...
+    var tickerC <-chan time.Time
+    if l.exportInterval > 0 {
+        ticker := time.NewTicker(l.exportInterval)
+        defer ticker.Stop()
+        tickerC = ticker.C
+    }
+    
+    for {
+        select {
+        case <-l.stopCh:
```
ErikPelli, line 138: Completely remove stopCh, we can close dataCh when we are done with the processing

ext/har/logger.go
```diff
@@ -0,0 +1,157 @@
...
+            l.drainChannel()
+            l.exportIfNeeded()
+            return
+        case entry := <-l.dataCh:
+            l.entries = append(l.entries, entry)
```
ErikPelli, line 143: entries must be a local slice in this exportLoop funcion, no need to set it as a field of the struct

ext/har/logger.go
```diff
@@ -0,0 +1,157 @@
...
+        tickerC = ticker.C
+    }
+    
+    for {
+        select {
+        case <-l.stopCh:
+            l.drainChannel()
+            l.exportIfNeeded()
+            return
+        case entry := <-l.dataCh:
```
ErikPelli, line 142:
```suggestion
        case entry, ok := <-l.dataCh:
```

When ok is false, you know that the data channel has been closed, instead of using stopCh

ext/har/logger.go
```diff
@@ -0,0 +1,157 @@
...
+// NewLogger creates a new HAR logger instance
+func NewLogger(exportFunc ExportFunc, opts ...LoggerOption) *Logger {
+    l := &Logger{
+        entries:        make([]Entry, 0),
+        captureContent: true,
+        exportFunc:     exportFunc,
+        exportThreshold: 100,    // Default threshold
+        exportInterval: 0,       // Default no interval
+        stopCh:         make(chan struct{}),
+        dataCh:         make(chan Entry, 1000),
```
ErikPelli, line 49:
```suggestion
        dataCh:         make(chan Entry), 
```

Use a normal channel, not a buffered channel, here it's enough

ext/har/logger.go
```diff
@@ -0,0 +1,157 @@
...
+
+// NewLogger creates a new HAR logger instance
+func NewLogger(exportFunc ExportFunc, opts ...LoggerOption) *Logger {
+    l := &Logger{
+        entries:        make([]Entry, 0),
+        captureContent: true,
+        exportFunc:     exportFunc,
+        exportThreshold: 100,    // Default threshold
+        exportInterval: 0,       // Default no interval
+        stopCh:         make(chan struct{}),
```
ErikPelli, line 48:
```suggestion
```

Remove this, as explained in another comment

ext/har/logger.go
```diff
@@ -0,0 +1,157 @@
...
+    return func(l *Logger) {
+        l.exportThreshold = threshold
+    }
+}
+
+// NewLogger creates a new HAR logger instance
+func NewLogger(exportFunc ExportFunc, opts ...LoggerOption) *Logger {
+    l := &Logger{
+        entries:        make([]Entry, 0),
+        captureContent: true,
```
ErikPelli, line 44: Why do you have this variable if its value is always true?

ext/har/logger.go
```diff
@@ -0,0 +1,157 @@
...
+func WithExportThreshold(threshold int) LoggerOption {
+    return func(l *Logger) {
+        l.exportThreshold = threshold
+    }
+}
+
+// NewLogger creates a new HAR logger instance
+func NewLogger(exportFunc ExportFunc, opts ...LoggerOption) *Logger {
+    l := &Logger{
+        entries:        make([]Entry, 0),
```
ErikPelli, line 43:
```suggestion
        entries:        make([]Entry, 0),
```
Default nil value for a slice is enough, no need to manually initialize it

ext/har/logger.go
```diff
@@ -0,0 +1,157 @@
...
+        },
+    }
+    entry.fillIPAddress(ctx.Req)
+    
+    select {
+    case l.dataCh <- entry:
+    default:
+        // Log or handle case where channel is full, just in case 
+        ctx.Proxy.Logger.Printf("Warning: HAR logger channel is full, dropping entry")
+    }
```
ErikPelli, line 97: No need to do that, just write to the channel

ext/har/logger.go
```diff
@@ -0,0 +1,157 @@
...
+    l.entries = append(l.entries, entry)
+    if l.exportThreshold > 0 && len(l.entries) >= l.exportThreshold {
+        l.exportFunc(l.entries)
+        l.entries = make([]Entry, 0)
+    }
+}
+
+func (l *Logger) exportIfNeeded() {
+    if len(l.entries) > 0 {
+        l.exportFunc(l.entries)
```
ErikPelli, line 112: Copy the slice data into another slice and spawn exportFunc() in another async goroutine

ext/har/logger.go
```diff
@@ -0,0 +1,157 @@
...
+        l.entries = make([]Entry, 0)
```
ErikPelli, line 113:
```suggestion
        l.entries = nil
```

ext/har/types.go
```diff
@@ -0,0 +1,356 @@
...
+    "io"
+    "net/http"
+    "net/url"
+    "mime"
+    "net"
+    "strings"
+    "time"
+    "github.com/elazarl/goproxy"
+)
+
```
ErikPelli, line 16:
```suggestion
```

Remove the extra space

ext/har/logger.go
```diff
@@ -0,0 +1,157 @@
...
+        defer ticker.Stop()
+        tickerC = ticker.C
+    }
+    
+    for {
+        select {
+        case <-l.stopCh:
+            l.drainChannel()
+            l.exportIfNeeded()
+            return
```
ErikPelli, line 141: Move this to dataCh when ok value is false.
In that case, the data channel is closed and you just pass the existing slice to the entries handler. No need to use a custom drainChannel or something like this.

ext/har/logger.go
ErikPelli, line 112: Consider to embed this logic inside exportLoop, they are just two if statements and by separating them, the code is harder to read

ext/har/logger.go
ErikPelli: With this I mean something like
```
var entries []SliceType
```

ext/har/logger_test.go
```diff
@@ -0,0 +1,187 @@
...
+
+            background := httptest.NewServer(ConstantHandler("hello world"))
+            defer background.Close()
+
+            proxyServer := createTestProxy(logger)
+            defer proxyServer.Close()
+
+            client := createProxyClient(proxyServer.URL)
+
+            req, err := http.NewRequest(tc.method, background.URL, strings.NewReader(tc.body))
```
ErikPelli, line 86: Use NewRequestWithContext
```suggestion
            req, err := http.NewRequestWithContext(context.Background(), tc.method, background.URL, strings.NewReader(tc.body))
```

ext/har/logger_test.go
```diff
@@ -0,0 +1,187 @@
...
+    defer background.Close()
+
+    proxyServer := createTestProxy(logger)
+    defer proxyServer.Close()
+
+    client := createProxyClient(proxyServer.URL)
+
+    // Send 3 requests
+    for i := 0; i < 3; i++ {
+        resp, err := client.Get(background.URL)
```
ErikPelli, line 176: Avoid client.Get, use client.Do instead, as we do in all the other tests

ext/har/logger_test.go
```diff
@@ -0,0 +1,187 @@
...
+    defer background.Close()
+
+    proxyServer := createTestProxy(logger)
+    defer proxyServer.Close()
+
+    client := createProxyClient(proxyServer.URL)
+
+    // Send 7 requests
+    for i := 0; i < 7; i++ {
+        resp, err := client.Get(background.URL)
```
ErikPelli, line 136: Avoid client.Get

ext/har/logger_test.go
```diff
@@ -0,0 +1,187 @@
...
+    
+    exportFunc := func(entries []Entry) {
+        entriesCopy := make([]Entry, len(entries))
+        copy(entriesCopy, entries)
+        exports = append(exports, entriesCopy)
+        t.Logf("Export occurred with %d entries", len(entries))
+        wg.Done()
+    }
+
+    logger := NewLogger(exportFunc, WithExportInterval(100*time.Millisecond))
```
ErikPelli, line 164: Avoid to use milliseconds, or this one will become a flaky test.
Use a suitable threshold instead, or find a deterministic way to solve this issue.

ext/har/types.go
```diff
@@ -0,0 +1,356 @@
...
+import (
+    "bytes"
+    "io"
+    "net/http"
+    "net/url"
+    "mime"
+    "net"
+    "strings"
+    "time"
+    "github.com/elazarl/goproxy"
```
ErikPelli, line 14:
```suggestion

    "github.com/elazarl/goproxy"
```

Separate stdlib from all the other libraries

ext/har/logger_test.go
```diff
@@ -0,0 +1,187 @@
...
+    "net/http"
+    "net/http/httptest"
+    "net/url"
+    "strings"
+    "testing"
+    "time"
+    "sync"
+    "github.com/elazarl/goproxy"
+    "github.com/stretchr/testify/assert"
+    "github.com/stretchr/testify/require"
```
ErikPelli, line 14:
```suggestion

    "github.com/elazarl/goproxy"
    "github.com/stretchr/testify/assert"
    "github.com/stretchr/testify/require"
```

ext/har/logger.go
```diff
@@ -0,0 +1,157 @@
+package har
+
+import (
+    "net/http"
+    "time"
+    "github.com/elazarl/goproxy"
```
ErikPelli, line 6:
```suggestion

    "github.com/elazarl/goproxy"
```

ext/har/types.go
```diff
@@ -0,0 +1,356 @@
...
+
+func parsePostData(ctx *goproxy.ProxyCtx, req *http.Request) *PostData {
+    harPostData := new(PostData)
+    
+    mediaType := parseMediaType(ctx, req.Header)
+    if mediaType == "" {
+        return nil
+    }
+    
+    harPostData.MimeType = mediaType
```
ErikPelli, line 154:
```suggestion
    mediaType := parseMediaType(ctx, req.Header)
    if mediaType == "" {
        return nil
    }
    
    harPostData = &PostData{
        MimeType: mediaType,
    }
```

ext/har/types.go
```diff
@@ -0,0 +1,356 @@
...
+			Pages:   make([]Page, 0, 10),
+			Entries: makeNewEntries(),
+		},
+	}
+	return har
+}
+
+func makeNewEntries() []Entry {
+    const startingEntrySize int = 1000
+	return make([]Entry, 0, startingEntrySize)
```
ErikPelli, line 48: Reformat these lines, there are too many spaces in the last

ext/har/logger.go
CameronBadman: @ErikPelli I can agree with that for ProccessEntry, however for exportIfNeeded, the same logic gets repeated 3 times,  would it be a better idea to put this as a closure function like this? 

```go 
func (l *Logger) exportLoop() {
    entries := make([]Entry, 0)
 
    exportIfNeeded := func() {
        if len(entries) > 0 {
            entriesCopy := make([]Entry, len(entries))
            copy(entriesCopy, entries)
            go l.exportFunc(entriesCopy)
            entries = nil
        }
    }
    
    // Rest of the function...
}


```

ext/har/types.go
CameronBadman: did you mean this here

```go
harPostData := &PostData{
        MimeType: mediaType,
    }
```

ErikPelli: yes!

ext/har/logger.go
CameronBadman: haven't done that part yet 

> Why are you attached with `make([]Entry, 0)`? The zero value is completely fine and you have to avoid `make` for an empty slice. When you want to reset it, set it to nil. A closure is ok, by the way!

CameronBadman: Also I genuinely haven't seen that before in programming, I didn't know you could append to a typed variable even if it is set to nil

ErikPelli: Don't worry, it's just an hint to improve your code! Appending to nil is great in Go because this prevents a lot of headaches!

ext/har/logger_test.go
CameronBadman: what do you mean by threshold? do you mean include a entry limit in that test? 

also wouldn't that test be deterministic because of the wg.Wait()?

ErikPelli: Let me explain it better, currently you're assuming that all the 3 requests will be exported in the same batch, but this can happen only if all of them will complete before 100 milliseconds, and this is a short time frame, so maybe sometimes this won't be true.

To fix this, I would increase it to 1 seconds, to have a great time margin and avoid cases like this.
Or, if you have any alternative proposal, you can say it too!

CameronBadman:
@ErikPelli Fixed up those errors I believe

ErikPelli:
@CameronBadman the comment that I left for the tests has been resolved but I don't see the requested changes in the branch

ext/har/logger_test.go
```diff
@@ -0,0 +1,195 @@
...
+
+func TestHarLoggerExportInterval(t *testing.T) {
+    var wg sync.WaitGroup
+    var exports [][]Entry
+    wg.Add(1) // Expect 1 export with all entries
+    
+    exportFunc := func(entries []Entry) {
+        entriesCopy := make([]Entry, len(entries))
+        copy(entriesCopy, entries)
+        exports = append(exports, entriesCopy)
```
ErikPelli, line 167: Why not directly use something like:
```
exports = append(exports, entries)
```
?

ext/har/logger_test.go
```diff
@@ -0,0 +1,195 @@
...
+        t.Logf("Export occurred with %d entries", len(entries))
+        wg.Done()
+    }
+
+    logger := NewLogger(exportFunc, WithExportInterval(100*time.Millisecond))
```
ErikPelli, line 172:
```suggestion
    logger := NewLogger(exportFunc, WithExportInterval(time.Second))
```

ext/har/logger_test.go
ErikPelli: Use `client.Do` instead, with a request built using `http.NewRequestWithContext()`

ext/har/logger_test.go
CameronBadman: that's copilot being funky and my vim not saving

ext/har/logger.go
```diff
@@ -0,0 +1,128 @@
...
+            entries = append(entries, entry)
+            if l.exportThreshold > 0 && len(entries) >= l.exportThreshold {
+                exportIfNeeded()
+            } 
+        case <-tickerC:
+            exportIfNeeded()
+        }
+    }
+}
+
```
ErikPelli, line 124:
```suggestion
```

ext/har/logger_test.go
```diff
@@ -0,0 +1,195 @@
...
+
+func TestLoggerThresholdExport(t *testing.T) {
+    var wg sync.WaitGroup
+    var exports [][]Entry
+    wg.Add(3) // Expect 3 exports (3,3,1)
+    
+    exportFunc := func(entries []Entry) {
+        entriesCopy := make([]Entry, len(entries))
+        copy(entriesCopy, entries)
+        exports = append(exports, entriesCopy)
```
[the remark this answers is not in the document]
CameronBadman: not entirely sure if I have solved this one

ext/har/logger_test.go
CameronBadman: running a copy gives further protection in the circumstance of future testing, like lets say modification to individual entries, its more about future proofing, but if you think it isn't worth I can remove it

ErikPelli: Yes please remove this here and in the other test, it's just a redundant copy

ext/har/logger_test.go
```diff
@@ -0,0 +1,222 @@
...
+
+    // Sort exports by length to ensure consistent checking
+    sort.Slice(exports, func(i, j int) bool {
+        return len(exports[i]) < len(exports[j])
+    })
+
+    require.Equal(t, 3, len(exports), "should have 3 export batches")
+    assert.Equal(t, 1, len(exports[0]), "last batch should have remainder")
+    assert.Equal(t, threshold, len(exports[1]), "first full batch should have threshold size")
+    assert.Equal(t, threshold, len(exports[2]), "second full batch should have threshold size")
```
ErikPelli, line 172: You need to check that two batches contain 3 entries and 1 batch contains 1 entry.
You're doing it wrong since you're relying on the index.
They can be in any order since we're using aynschornous goroutines for export func.

ext/har/logger_test.go
```diff
@@ -0,0 +1,222 @@
...
+
+func TestHarLoggerExportInterval(t *testing.T) {
+    var wg sync.WaitGroup
+    var mu sync.Mutex
+    var exports [][]Entry
+    wg.Add(1) // Expect 1 export with all entries
+    
+    exportFunc := func(entries []Entry) {
+        entriesCopy := make([]Entry, len(entries))
+        copy(entriesCopy, entries)
```
ErikPelli, line 183: Remove copy

ext/har/logger_test.go
```diff
@@ -0,0 +1,222 @@
...
+    var exports [][]Entry
+    var mu sync.Mutex
+    wg.Add(3) // Expect 3 exports (3,3,1)
+    
+    exportFunc := func(entries []Entry) {
+        entriesCopy := make([]Entry, len(entries))
+        copy(entriesCopy, entries)
+        
+        mu.Lock()
+        exports = append(exports, entriesCopy)
```
ErikPelli, line 129:
```suggestion
        mu.Lock()
        exports = append(exports, entries)
```

ext/har/logger_test.go
CameronBadman: @ErikPelli sorry won't the sort.Slice put them in the correct order?

ErikPelli: Uuups, I missed the sort. My bad.

ErikPelli:
Thanks! LGTM. Tomorrow I'll do one final check and, if it will be ok, I will merge this pull request

ErikPelli:
Thanks!

PR #615, opened 2025-01-07
Inherit `go` version from `go.mod` file

jeffwidman:
This way it's one less place the version needs to be bumped.

Whenever the minimum version in `go.mod` is bumped, that will be immediately reflected here.


ErikPelli:
The setup-go stage is failing, uhmmm. Maybe the path to go.mod must be absolute? Any idea?

jeffwidman:
Ahh, good catch!

PR #621, opened 2025-01-07
redundant names are redundant

jeffwidman:
The actions names here are self-explanatory, no need to name them.

Plus the "checkout" name was slightly incorrect, as it checks out the full repo, and has nothing to do with the go modules directory.


ErikPelli (approved):

ErikPelli:
Since this is only a styling issue and has no impact, I will merge this pull request along with another one, so I'll release only one version for both of them

ErikPelli:
Thanks!

Issue #633, opened 2025-01-20
Context Done Not Propagated in Goproxy Extension Limiter During HTTPS/CONNECT


alessiodallapiazza:
The Goproxy extension limiter fails to propagate the context Done signal correctly during HTTPS/CONNECT requests. This causes the extension to block SSL connections, resulting in undesired behavior. Steps to reproduce and potential fixes would be appreciated.

Here's an example of a minimal but sufficient piece of code to reproduce the issue:
```go
package main

import (
	"log"
	"net/http"
	"net/url"
	"time"
	"crypto/tls"
	"context"

	"github.com/elazarl/goproxy"
)

const maxConcurrentRequests = 1

func concurrentRequests(limit int) goproxy.ReqHandler {
	if limit <= 0 {
		return goproxy.FuncReqHandler(func(req *http.Request, ctx *goproxy.ProxyCtx) (*http.Request, *http.Response) {
			return req, nil
		})
	}

	limitation := make(chan struct{}, limit)
	return goproxy.FuncReqHandler(func(req *http.Request, ctx *goproxy.ProxyCtx) (*http.Request, *http.Response) {
		log.Printf("Request received: %s %s - queue size: %d", req.Method, req.URL.String(), len(limitation))

		reqCtx, cancel := context.WithTimeout(req.Context(), 10*time.Second)
		req = req.WithContext(reqCtx)

		select {
		case limitation <- struct{}{}:
			go func() {
				select {
				case <-reqCtx.Done():
					log.Printf("Request finished or timed out: %s %s", req.Method, req.URL.String())
					<-limitation
					cancel()
				}
			}()
			return req, nil
		case <-reqCtx.Done():
			log.Printf("Request timed out while waiting: %s %s", req.Method, req.URL.String())
			cancel()
			return req, goproxy.NewResponse(req, goproxy.ContentTypeText, http.StatusServiceUnavailable, "Request timed out while waiting for available slot")
		}
	})
}

func startProxyServer() {
	proxy := goproxy.NewProxyHttpServer()
	proxy.OnRequest().HandleConnect(goproxy.AlwaysMitm)
	proxy.Verbose = false

	tr := &http.Transport{
		TLSClientConfig: &tls.Config{
			InsecureSkipVerify: true,
		},
		DisableKeepAlives: true,
		ForceAttemptHTTP2: false,
	}
	proxy.Tr = tr

	proxy.OnRequest().Do(concurrentRequests(maxConcurrentRequests))

[...]

PR #634, opened 2025-01-20
Cancel MITM request context when its handling finishes

ErikPelli:
Fixes https://github.com/elazarl/goproxy/issues/633.
Add Go context cancel to a custom parsed request, since we handle it.

### DIFF
--- a/https.go
+++ b/https.go
@@ -329,70 +329,75 @@ func (proxy *ProxyHttpServer) handleHttps(w http.ResponseWriter, r *http.Request
 				// information URL in the context when does HTTPS MITM
 				ctx.Req = req
 
-				req, resp := proxy.filterRequest(req, ctx)
-				if resp == nil {
-					if req.Method == "PRI" {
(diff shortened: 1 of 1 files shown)


ErikPelli:
@alessiodallapiazza I tested your example script with this release and it works perfectly.
If it's all good for you, I think I can merge this pull request as a stable release, since we're also passing the test pipeline without issues.
What do you think?

alessiodallapiazza:
I’ve run some tests as well, and everything looks good on my end.  👍

ErikPelli:
I don't think we need to introduce new tests for every single change we made, do you think they are necessary for this one?

thisislawatts:
This specific change was made to address a specific issue #633. So I guess if it was a significant enough bug to warrant a change it also seems worth capturing the desired behaviour in an automated test. Otherwise, I am curious how you would avoid introducing regressions or reoccurrences of the same bug?

Issue #630, opened 2025-01-17
Use directly HTTP transport to establish websockets


ErikPelli:
Instead of using a custom handler for websockets, with a "101 Switching Protocol" response, the same TCP connection is reused for websockets.
According to go standard library ([*http.Response](https://go.dev/src/net/http/response.go)):
```
// As of Go 1.12, the Body will also implement io.Writer
// on a successful "101 Switching Protocols" response,
// as used by WebSockets and HTTP/2's "h2c" mode.
```

So we can directly reuse it as a TCP connection.

ErikPelli:
Call response handlers after 101 but before the raw data mode, so the user can insert its wrapper in the body

PR #637, opened 2025-01-21
Handle Websocket using standard library

ErikPelli:
Fixes https://github.com/elazarl/goproxy/issues/630.
Instead of relying on custom code to handshake WebSockets separately, just rely on the standard library response Body, as explained in the documentation.
This will also use a cascadeproxy configuration by default, without the need to setting a custom per-request Dialer, when the custom transport is configured with a proxy URL.

I also update the websocket example to use a more updated WebSocket client library.


ErikPelli:
Sure! Take your time

ErikPelli:
@elazarl have you had any chance to look at this PR? Can I merge it?

PR #639, opened 2025-01-21
Add telegram group community

ErikPelli:
I created a Telegram group to have a quick chat with the users about issues and features, more direct than issues and pull requests.
What do you think? @elazarl


elazarl:
I created a gitter account back then, and it didn't catch, but I guess times have changed, and it worth a shot!
I know some people uses discord, but Telegram seems viable.
I'll join the channel.

PR #669, opened 2025-03-26
Fix http connect handling

jessemalone:
This PR fixes an issue in the Connect handler the breaks forwarded HTTP requests. The getTargetSiteConnection should pull the target port from the request (host) if possible. Otherwise plain http request are sent to servers that expect TLS and will fail.


jessemalone:
This was *supposed* to be against the Windscribe fork, not upstream.

Issue #733, opened 2026-01-08
Issue with keep alive


ErikPelli:
@elazarl there are some hidden ghosts from 2017 👻 
Someone signaled me privately that keep alive are basically not working, after almost 10 years.
Basically in https://github.com/elazarl/goproxy/pull/211 & https://github.com/elazarl/goproxy/issues/209 someone added a Connection: close to every request intercepted by the proxy, and so the browser that uses the proxy is closing every single connection.
This means that for every single request, a new TCP connection is opened, affecting the performances by a lot.
In your opinion should we remove this at all to comply with the standard request processing (I would never expect a proxy to add this header if the origin didn't include it)? The only cons of changing this is that maybe someone is expecting that the proxy will close every connection, but this seems strange.
Maybe can we add an entry to the proxy configuration and the default is to not close them, but if it's true, they are automatically closed like they currently are?
What do you think?

elazarl:
I think we should change this behavior, and not worry about breaking old behavors.

What is the reason for the original change?

mikel-jason:
>Should we remove this at all and make the keep alive working again?

We use this project for a custom authorization proxy inside k8s (thanks btw <3 ) and now got the requirement to onboard the proxy to our Istio service mesh. I learnt just yesterday that it won't work because the sidecar tries to establish persistent connections but apparently fails to do so. Most likely, that's on the behavior described in this issue. More details: https://github.com/istio/istio/issues/55138#issuecomment-2666855044

Looking forward to this change, either default or configurable.

PR #734, opened 2026-01-12
Fix keep alive logic and replace legacy response write logic

ErikPelli:
Issue https://github.com/elazarl/goproxy/issues/733 explains a problem that we currently have with the keep alive in the connection between the proxy client and the proxy server.
In the current implementation, if the origin sends a response, the proxy server is adding a `Connection: close` header that is preventing the keep alive to work correctly.
For this exceptional case, in the issue we decided to "break the compatibility" because this is clearly a bug that is preventing the proxy server from following the standard.
We are currently manually editing the origin response and this is clearly not ok, since keep alive is an important optimization in HTTP/1.1.

There are 2 issues that i fixed in this pull request:
1. The Connection: close header added by the proxy server
2. If the origin is sending a body, and the body is not modified by the proxy response handlers, the proxy server is calling rawClientTls.Close(), closing the TLS connection, I just removed it, so the proxy client can reuse the same connection for other requests

### DIFF
--- a/proxy_test.go
+++ b/proxy_test.go
@@ -7,11 +7,13 @@ import (
 	"crypto/tls"
 	"crypto/x509"
 	"encoding/base64"
+	"fmt"
 	"io"
 	"log"
 	"net"
 	"net/http"
 	"net/http/httptest"
+	"net/http/httptrace"
 	"net/url"
 	"os"
 	"os/exec"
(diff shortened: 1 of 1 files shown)


ErikPelli:
@elazarl I went too far with this issue and I decided to remove all the crappy code for the chunk and the response writing and just use (*http.Response).Write(), to avoid any logic error and wrong assumptions.
Now it looks much more cleaner.
Additionally, I added a test for websockets.
The pipeline is failing due to linting issues, I'll have a look at it later.

Please review it and let me know if it's ok for you.

Issue #740, opened 2026-01-24
Regression: Client connection open forever if response doesn't have "Content-Length" header


nihiloid:
Example: goproxy accepting HTTP/1.1 client connections and MITMing on 127.0.0.1:8080 with HTTP/2 server support.
```
curl 'https://example.com/' \
  -H 'accept: */*' \
  -x http://127.0.0.1:8080 -k
```
No `content-length` in HTTP/2 server response.

with **v1.8.0** returns no transfer encoding header, client connection is opened forever.

with **v1.7.2** returns `Transfer-Encoding: chunked` header, client connection closed as expected.

ErikPelli:
Please review https://github.com/elazarl/goproxy/pull/742

PR #742, opened 2026-01-30
Use chunked response when http2 content length is missing

ErikPelli:
https://github.com/elazarl/goproxy/pull/720 by @Skn0tt introduced a major bug in the codebase, as signaled by @nihiloid in this issue: https://github.com/elazarl/goproxy/issues/740.
I successfully replicated with a test the behaviour and in fact the issue was right.
When an HTTP/2 server doesn't respond with a Content-Length header, the current implementation doesn't set the chunked response and the HTTP client becomes broken, waiting undefinitely.
In this pull request I'm adding the chunked response also when the Content-Length header is empty (""), resolving this edge case that is very important to address. I also added a test to prevent this in the future.
In fact pull request #720 (partially) broke the compatibility with HTTP/2.

@nihiloid @Skn0tt please review this and let me know if everything is ok.
@Skn0tt I don't know if this change affects Playwright, let me know.

### DIFF
--- a/https.go
+++ b/https.go
@@ -380,8 +380,12 @@ func (proxy *ProxyHttpServer) handleHttps(w http.ResponseWriter, r *http.Request
 					resp = proxy.filterResponse(resp, ctx)
 					bodyModified := resp.Body != origBody
 					defer resp.Body.Close()
-					if bodyModified || resp.Header.Get("Content-Length") == "" {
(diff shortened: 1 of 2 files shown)


nihiloid:
Tested on my use case, the issue has been fixed. Thanks @ErikPelli

Skn0tt (approved):
Nice find!

PR #755, opened 2026-03-05
Fix MITM responses leaking upstream HTTP/2 protocol version

robmry:
Firstly, thank you for this project!

When bumping from 1.7.2 to 1.8.2 I hit an issue with MITM-ing a connection between an HTTP/1 client and an HTTP/2 server.

I'm afraid the rest of this is from an AI - please do let me know if anything needs de-slop-ing, but I think it's sane.

---

When the proxy connects to an upstream server over HTTP/2, the `http.Response` has Proto "HTTP/2.0".

Commit 78c76be5 ("Fix keep alive logic and replace legacy response write logic") switched from manually writing "HTTP/1.1" status lines to using resp.Write(), and commit b343a9ac ("Merge HTTPMitmConnect and MitmConnect actions") unified both MITM code paths.

Together, these changes cause the upstream protocol version to be preserved verbatim, writing "HTTP/2.0 200 OK" to the MITM'd client which speaks HTTP/1.1 over the hijacked connection. This causes clients to reject or reset the connection.

Normalize resp.Proto to HTTP/1.1 before calling resp.Write() in the MITM handler, since in this code path the client-facing connection is always HTTP/1.1.

### DIFF
--- a/https.go
+++ b/https.go
@@ -349,6 +349,14 @@ func (proxy *ProxyHttpServer) handleHttps(w http.ResponseWriter, r *http.Request
 						resp.TransferEncoding = []string{"chunked"}
 					}
 
+					// The MITM'd client speaks HTTP/1.1, but the upstream
+					// response may have been received over HTTP/2. Normalize
+					// the protocol version so resp.Write() produces a valid
+					// HTTP/1.1 status line.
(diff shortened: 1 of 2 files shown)


ErikPelli:
Thanks for your contribution!

PR #761, opened 2026-04-09
http2: set ServerName for upstream TLS client

germanescobar:
Hi, we are using goproxy for a sandbox platform that we are building and found this issue that we had to patch in our project, so I asked Codex to help me open this PR:

This fixes an HTTP/2 MITM failure in `H2Transport.RoundTrip()` when the TLS config used for the upstream client connection does not already include `ServerName`.

Before this change, the H2 path could fail with:

`tls: either ServerName or InsecureSkipVerify must be specified in the tls.Config`

The root issue is that `RoundTrip()` reuses `r.TLSConfig` as a client TLS config, but that config is often created for the server-side MITM role and may not yet be valid for upstream client verification.

This change:
- clones the provided TLS config instead of mutating it in place
- restricts ALPN on the cloned config to `h2`
- populates `ServerName` from the upstream host when it is missing
- verifies the hostname against that resolved server name

I hit this while testing Git smart-HTTP over HTTPS with downstream HTTP/2 enabled through the proxy.

### DIFF
--- a/h2.go
+++ b/h2.go
@@ -33,15 +33,28 @@ func (r *H2Transport) RoundTrip(_ *http.Request) (*http.Response, error) {
 	if !strings.Contains(raddr, ":") {
 		raddr += ":443"
 	}
+	serverName := r.Host
+	if host, _, err := net.SplitHostPort(raddr); err == nil {
+		serverName = host
+	}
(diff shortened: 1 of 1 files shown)


ErikPelli:

PR #767, opened 2026-04-10
http2: add a generic MITM header rewrite hook

### LINKED ISSUE #766: HTTP/2 MITM path needs a generic request header rewrite hook
## Summary

We ended up needing a small local fork of `goproxy` to make HTTP/2 MITM request mutation work for our production use case.

The core gap is that the HTTP/2 MITM path currently bypasses the normal request mutation flow and raw-proxies H2 frames, so code that can rewrite HTTP/1.1 MITM requests cannot apply equivalent mutations to downstream HTTP/2 requests.

In our case, this surfaced as credentials embedded in URL userinfo / `Authorization: Basic ...` being rewritten correctly on the HTTP/1.1 MITM path, but not on the downstream HTTP/2 MITM path.

## What we observed

Our use case is a per-sandbox outbound MITM proxy that substitutes placeholders in outbound request data before forwarding upstream.

This worked on the normal MITM path, but failed on the downstream H2 path because the H2 implementation in `https.go` / `h2.go` forwards client->server HEADERS frames directly and does not expose a generic header-mutation hook.

The result was:

- downstream client speaks HTTP/2 to the proxy
- client sends URL credentials or `Authorization: Basic ...`
- upstream request is forwarded without our application-level mutation logic being applied
- auth fails upstream even though the same flow works on the HTTP/1.1 path

## How we solved it locally

We made a local fork change with two generic pieces:

1. In `https.go`, in the `req.Method == "PRI"` branch, wire an optional header-rewrite callback into `H2Transport`.
2. In `h2.go`, decode the client->server HPACK header block, expose the decoded fields to the callback, then re-encode and forward the modified block.

Conceptually, the added hook looks like:

```go
// in H2Transport
RewriteHeaders func([]H2HeaderField) []H2HeaderField
```

and in the H2 MITM path:

- decode HEADERS/CONTINUATION into header fields
- call `RewriteHeaders(...)` if present
- re-encode the modified header block
- write the new HEADERS/CONTINUATION frames upstream

This let our application keep all product-specific logic outside the fork while only carrying a small generic extension point inside the proxy.

## Why this would help upstream

This seems generally useful beyond our product:

- auth header rewriting
- host/path normalization
- instrumentation / policy enforcement on H2 MITM requests
- keeping app-specific logic outside a growing proxy fork

Right now, the HTTP/2 MITM path is much harder to customize than the HTTP/1.1 path even though both are MITM request-processing paths.

## Scope suggestion

I’m not proposing any sandbox-specific behavior upstream.

The request is for a generic extension point on the downstream HTTP/2 MITM path so callers can mutate decoded request headers before they are forwarded upstream.

If helpful, I can open a PR with the minimal generic version of the hook we used locally.

germanescobar:
## Summary
- add a generic HTTP/2 request header rewrite hook for the MITM path
- wire the hook from the existing `ctx.RoundTripper` customization surface
- add a focused unit test covering HPACK decode/rewrite/re-encode for request headers

## Why
The downstream HTTP/2 MITM path currently raw-proxies HEADERS frames, so callers can customize HTTP/1.1 MITM requests but cannot apply equivalent header mutation to HTTP/2 requests.

This adds a small generic extension point:
- `H2HeaderField`
- `H2HeaderRewriter`
- `H2Transport.RewriteHeaders`

When a custom `ctx.RoundTripper` also implements `H2HeaderRewriter`, the HTTP/2 MITM path decodes the client->server HPACK header block, lets the caller mutate it, then re-encodes and forwards it upstream.

Closes #766.

## Testing
- `go test ./...`

### DIFF
--- a/h2.go
+++ b/h2.go
@@ -2,6 +2,7 @@ package goproxy
 
 import (
 	"bufio"
+	"bytes"
 	"context"
 	"crypto/tls"
 	"errors"
@@ -11,6 +12,7 @@ import (
 	"strings"
 
 	"golang.org/x/net/http2"
+	"golang.org/x/net/http2/hpack"
 )
 
 var ErrInvalidH2Frame = errors.New("invalid H2 frame")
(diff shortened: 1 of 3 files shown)


ErikPelli:

PR #769, opened 2026-04-11
Emit declared response trailers

dimitarvdimitrov:
> prompted codex into fixing this, the fix looks good to me, but open to feedback :)

## Problem 

`handleHttp` was forwarding the `Trailer` header declaration but never writing the trailer values onto the downstream `http.ResponseWriter`.

That meant clients could see `Trailer: Server-Timing` in the response headers, but the actual trailing `Server-Timing` fields were never sent.

## What this PR does

This updates the HTTP response write path to actually emit declared response trailers to downstream clients.

- drop `Content-Length` when trailers are declared so HTTP/1.1 can use chunked framing (otherwise no trailers are expected after `Content-Length` number of bytes). 
  - HTTP/2 doesn't have behave like this and this change means that even for HTTP/2 we use chunked when there are trailers and we would have otherwise used regular encoding
- copy `resp.Trailer` onto the downstream `ResponseWriter` after the body finishes streaming

### DIFF
--- a/http.go
+++ b/http.go
@@ -58,10 +58,16 @@ func (proxy *ProxyHttpServer) handleHttp(w http.ResponseWriter, r *http.Request)
 	// We keep the original body to remove the header only if things changed.
 	// This will prevent problems with HEAD requests where there's no body, yet,
 	// the Content-Length header should be set.
-	if origBody != resp.Body {
+	announcedTrailers := declaredTrailers(resp)
+	if origBody != resp.Body || len(announcedTrailers) > 0 {
 		resp.Header.Del("Content-Length")
 	}
 	copyHeaders(w.Header(), resp.Header, proxy.KeepDestinationHeaders)
+	for trailer := range announcedTrailers {
+		if !headerDeclaresTrailer(resp.Header, trailer) {
+			w.Header().Add("Trailer", trailer)
+		}
+	}
 	w.WriteHeader(resp.StatusCode)
 
 	if isWebSocketHandshake(resp.Header) {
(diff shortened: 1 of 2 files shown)


http.go
```diff
@@ -58,10 +58,16 @@ func (proxy *ProxyHttpServer) handleHttp(w http.ResponseWriter, r *http.Request)
...
 	// This will prevent problems with HEAD requests where there's no body, yet,
 	// the Content-Length header should be set.
-	if origBody != resp.Body {
+	announcedTrailers := declaredTrailers(resp)
+	if origBody != resp.Body || len(announcedTrailers) > 0 {
 		resp.Header.Del("Content-Length")
 	}
 	copyHeaders(w.Header(), resp.Header, proxy.KeepDestinationHeaders)
+	for trailer := range announcedTrailers {
+		if !headerDeclaresTrailer(resp.Header, trailer) {
```
obarisk, line 67: what's the average number of trailers?
headerDeclaresTrailer will be called many times.

if we often with only a few. then it's okay.

dimitarvdimitrov: the only one i'm aware of is Server-Timing. I can make it more efficient and build the `declaredTrailersFromHeader` only once instead of for every map 👍

ErikPelli:
This seems the same of https://github.com/elazarl/goproxy/pull/773 but with more duplicated and redundant code, maybe I will evaluate it first

dimitarvdimitrov:
oh, amazing, it's already fixed in #773. yeah my code was a bit more sloppy. Thanks @0b01 !

PR #771, opened 2026-04-15
fix: append headers instead of overwriting when keepDestHeaders is true

### LINKED ISSUE #752: Set-Cookie - any non unique header name in destination is overwritten by source
This is same issue as https://github.com/elazarl/goproxy/issues/205

Problem is here:

```
func copyHeaders(dst, src http.Header, keepDestHeaders bool) {
	if !keepDestHeaders {
		for k := range dst {
			dst.Del(k)
		}
	}
	for k, vs := range src {
		// direct assignment to avoid canonicalization
		dst[k] = append([]string(nil), vs...)
	}
}
```

In case source - backend adds Set-Cookie header it overwrites Set-Cookie header already set in destination altough in fact it is different cookie (but cookie name is same Set-Cookie) and same situation will be for any headers which don't have unique header names, so probably better would be, in case keepDestHeaders is enabled to just append to dst instead of overwriting by `[]string(nil) and then appending`

Yanhu007:
Fixes #752

## Problem

`copyHeaders` with `keepDestHeaders=true` overwrites destination header values instead of appending:

```go
dst[k] = append([]string(nil), vs...)
```

This drops all original destination values for non-unique headers like `Set-Cookie`. When a backend adds its own `Set-Cookie` header and the proxy response handler has already set a different `Set-Cookie`, the original is lost.

## Fix

When `keepDestHeaders` is true, append source values to existing destination values:

```go
if keepDestHeaders {
    dst[k] = append(dst[k], vs...)
} else {
    dst[k] = append([]string(nil), vs...)
}
```

This preserves all existing destination headers while adding source headers, correctly handling multi-value headers like `Set-Cookie`.

### DIFF
--- a/proxy.go
+++ b/proxy.go
@@ -62,8 +62,14 @@ func copyHeaders(dst, src http.Header, keepDestHeaders bool) {
 		}
 	}
 	for k, vs := range src {
-		// direct assignment to avoid canonicalization
-		dst[k] = append([]string(nil), vs...)
+		if keepDestHeaders {
+			// Append to existing values to preserve non-unique headers
+			// like Set-Cookie.
+			dst[k] = append(dst[k], vs...)
+		} else {
+			// direct assignment to avoid canonicalization
+			dst[k] = append([]string(nil), vs...)
+		}
 	}
 }
 
--- a/proxy.go
+++ b/proxy.go
@@ -58,17 +58,13 @@ var hasPort = regexp.MustCompile(`:\d+$`)
 func copyHeaders(dst, src http.Header, keepDestHeaders bool) {
 	if !keepDestHeaders {
 		for k := range dst {
-			dst.Del(k)
+			delete(dst, k)
 		}
 	}
 	for k, vs := range src {
 		if keepDestHeaders {
-			// Append to existing values to preserve non-unique headers
-			// like Set-Cookie.
-			dst[k] = append(dst[k], vs...)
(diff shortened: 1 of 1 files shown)


ErikPelli:
As I said in the issue, this change is useless since in the response handler you can just add a value to an header, and it works.
Here is an example:
```
func TestCookieMultipleValues(t *testing.T) {
	// target server
	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		w.Header().Set("Test", "1")
	}))
	defer srv.Close()

	// proxy server
	proxy.OnResponse().DoFunc(func(resp *http.Response, ctx *goproxy.ProxyCtx) *http.Response {
		resp.Header.Add("Test", "2")
		return resp
	})
	proxySrv := httptest.NewServer(proxy)
	defer proxySrv.Close()

	// send request
	client := &http.Client{}
	client.Transport = &http.Transport{
		Proxy: func(req *http.Request) (*url.URL, error) {
			return url.Parse(proxySrv.URL)
		},
	}
	req, _ := http.NewRequestWithContext(context.Background(), http.MethodGet, srv.URL, nil)
	resp, _ := client.Do(req)

	assert.Len(t, resp.Header["Test"], 2)
	assert.Contains(t, resp.Header["Test"], "1")
	assert.Contains(t, resp.Header["Test"], "2")
}
```

PR #772, opened 2026-04-17
h2: terminate HTTP/2 sessions instead of forwarding raw frames

0b01:
## Summary

This PR replaces the frame-level HTTP/2 forwarder (`H2Transport`) with a
proper HTTP/2 termination layer. Each HTTP/2 stream is decoded into a
standard `*http.Request` and dispatched through the existing
`filterRequest` → `RoundTrip` → `filterResponse` pipeline, exactly like
HTTP/1.1.

## Why

The previous `H2Transport` copied frames between the client's and
upstream's `http2.Framer`. While this worked for blind passthrough, it
was structurally incompatible with goproxy's value proposition:

- Handlers registered via `OnRequest` / `OnResponse` never ran for H2
  traffic — the proxy never had decoded `*http.Request` objects.
- `:path`, `:method`, headers, and bodies were invisible to user code,
  so URL/header/body inspection or rewriting was impossible.
- gRPC traffic (HTTP/2 with paths like `/pkg.Service/Method`) was
  particularly affected: the only thing observable was the destination
  host, which is far too coarse for any kind of per-RPC policy.

After this change H2 traffic is a first-class citizen of the filter
chain.

### Core
- **`h2.go`**: `H2Transport` and the per-frame forwarding code are
  removed. New `serveH2` adapts the decrypted client conn into a
  `net.Conn` and drives it with `http2.Server.ServeConn`. The handler
  rewrites each request to look like an absolute-URL proxy request and
  hands it to the standard `handleHttp` path.
- **`https.go`**: The `PRI` method is now handled *before* the filter
  chain (it is the H2 connection preface, not a real request). When
  `AllowHTTP2` is true and the user's TLS config does not pin
  `NextProtos`, `h2`/`http/1.1` are advertised automatically so that
  H2 clients can negotiate without requiring callers to update their
  `TLSConfig` callback.
- **`http.go`**: New internal `handleHttpWithParent` accepts an outer
  `*ProxyCtx` and seeds `UserData` and `RoundTripper` for each
  per-stream context. This matches the semantics already used by the
  HTTP/1.1 path and prevents H2 streams from silently losing context
  set by an `OnConnect` handler.

### Lifecycle / correctness
- `h2Conn` now delegates `Close`, `SetDeadline`, `SetReadDeadline`,
  and `SetWriteDeadline` to the underlying connection. The previous
  no-op implementations meant `http2.Server` could neither enforce
  idle/read/write timeouts nor cleanly tear the connection down on
  GOAWAY or error.
- The original client `RemoteAddr` from the outer CONNECT is now
  propagated to each decoded H2 `*http.Request`.

### Tests
New `TestH2Mitm*` cases cover:

- Passthrough (`/test-path` echoed end-to-end, `Proto == HTTP/2.0`)
- Per-stream request filtering (`/blocked` returns 403, `/allowed`
  reaches upstream)
- Per-stream response filtering (proxy adds a header, upstream's
  header is preserved)
- Request header rewriting
- URL rewriting
- Response body replacement
- Multiple streams sharing one HTTP/2 connection through the proxy

A small `h2TestEnv` helper bundles the proxy, an in-process H2
upstream, a CONNECT-tunneled TLS conn, and a pinned
`http2.Transport`-based client to keep each test ~5 lines.

## Backwards compatibility

- `AllowHTTP2 = false` (the default) is unchanged: H2 connections are
  rejected exactly as before.
- Existing HTTP/1.1 behavior is unchanged.
- Callers that already set `tls.Config.NextProtos` keep full control —
  the auto-injection only kicks in when they leave it empty.
- The exported `H2Transport` type and `ErrInvalidH2Frame` are removed.
  These were only useful in conjunction with the previous forwarder
  and have no migration path because the new design does not expose
  raw frames; if anyone depends on them, they should not enable
  `AllowHTTP2` and should keep H2 as an opaque CONNECT tunnel.

## How tested

```
go test ./...
```

### DIFF
--- a/h2.go
+++ b/h2.go
@@ -1,208 +1,68 @@
 package goproxy
 
 import (
-	"bufio"
-	"context"
-	"crypto/tls"
-	"errors"
 	"io"
 	"net"
 	"net/http"
 	"strings"
+	"time"
 
 	"golang.org/x/net/http2"
 )
 
-var ErrInvalidH2Frame = errors.New("invalid H2 frame")
-
-// H2Transport is an implementation of RoundTripper that abstracts an entire
-// HTTP/2 session, sending all client frames to the server and responses back
-// to the client.
-type H2Transport struct {
-	ClientReader io.Reader
-	ClientWriter io.Writer
-	TLSConfig    *tls.Config
-	Host         string
+// h2Conn adapts an io.Reader/io.Writer pair into a net.Conn for
+// http2.Server.ServeConn.
+type h2Conn struct {
+	r    io.Reader
+	w    io.Writer
+	conn net.Conn
 }
 
-// RoundTrip executes an HTTP/2 session (including all contained streams).
-// The request and response are ignored but any error encountered during the
-// proxying from the session is returned as a result of the invocation.
-func (r *H2Transport) RoundTrip(_ *http.Request) (*http.Response, error) {
-	raddr := r.Host
-	if !strings.Contains(raddr, ":") {
-		raddr += ":443"
-	}
-	rawServerTLS, err := dial("tcp", raddr)
-	if err != nil {
-		return nil, err
-	}
-	defer rawServerTLS.Close()
-	// Ensure that we only advertise HTTP/2 as the accepted protocol.
-	r.TLSConfig.NextProtos = []string{http2.NextProtoTLS}
-	// Initiate TLS and check remote host name against certificate.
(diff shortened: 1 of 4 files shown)


ErikPelli:
This part is very critical, I will need time to review this and I will develop my own version with the best practices, I need to be 100% sure that nothing will break

ErikPelli:

PR #773, opened 2026-04-17
http: forward upstream response trailers

0b01:
When the upstream responds with trailers (HTTP/1.1 chunked Trailer block or HTTP/2 trailing HEADERS frame, as gRPC requires), goproxy was silently discarding them. The grpc-go client then surfaced the missing grpc-status as:

    rpc error: code = Internal desc = server closed the stream without sending trailers

`handleHttp` now mirrors net/http/httputil.ReverseProxy's trailer logic:

  1. Pre-announce upstream trailer keys via the "Trailer" header before WriteHeader. This is what makes net/http commit to chunked encoding (h1) or a trailing HEADERS frame (h2); without it, short bodies get a Content-Length and trailers are dropped on the floor.
  2. After body copy + Body.Close (which is what populates resp.Trailer for h2 trailing-HEADERS), force a Flush so the server can no longer buffer to a fixed-length response.
  3. Forward values: under unprefixed keys when every trailer was pre-announced (the gRPC-over-h1-upstream case), otherwise via http.TrailerPrefix so dynamically-added trailers also get through.

### DIFF
--- a/http.go
+++ b/http.go
@@ -62,6 +62,20 @@ func (proxy *ProxyHttpServer) handleHttp(w http.ResponseWriter, r *http.Request)
 		resp.Header.Del("Content-Length")
 	}
 	copyHeaders(w.Header(), resp.Header, proxy.KeepDestinationHeaders)
+
+	// Announce trailers known at this point (HTTP/1.1 with pre-announced
+	// Trailer header). Setting "Trailer" before WriteHeader makes
+	// http.Server commit to chunked encoding (h1) or a trailing HEADERS
+	// frame (h2), which is required for any trailers to be forwarded.
+	// Mirrors net/http/httputil.ReverseProxy.
+	announcedTrailers := len(resp.Trailer)
+	if announcedTrailers > 0 {
+		trailerKeys := make([]string, 0, announcedTrailers)
+		for k := range resp.Trailer {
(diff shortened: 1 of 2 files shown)


ErikPelli:
@0b01 this seems to be a good change, but you have applied it only to the http.go file, which means that it won't be used with HTTPS MITM, but only with plain HTTP.
The same change is needed for HTTPS or for it there is no need to change anything and the trailer already works?

0b01:
Hey @ErikPelli, good question.

HTTPS MITM doesn't need this. That [path]( https://github.com/0b01/goproxy/blob/31d64677f6c73622f00fec23a344314360934f9b/https.go#L389) uses [net/http/response.Write](https://cs.opensource.google/go/go/+/master:src/net/http/response.go;l=242?q=response.go&ss=go%2Fgo) which already writes the Trailer field. The bug only surfaced in conjunction with #772, specifically [this line](https://github.com/elazarl/goproxy/pull/772/changes#diff-0719e4fc9631720b90dde571c371e56f0ceaf094218f15611babe52e96e99810R65) which routes the MITM'd HTTP/2 stream through the `http` handler. gRPC requires Trailer to work.

ErikPelli:
Thanks!

Issue #732, opened 2025-12-29
Incorrect Transfer-Encoding set for 304 responses


cal-pratt:
Using version `1.7.2` I see that 304 responses are being set with "Transfer-Encoding" set to "chunked". The 304 responses appear to have no body (which seems correct), however, the response headers still list a "chunked" encoding. This is causing `pip` installs to timeout when using a MITM setup (timing out waiting for the chunked response). Code in question seems to be

https://github.com/elazarl/goproxy/blob/09a5680351ad4b2271bce85eda4414b1afc6efd0/https.go#L395-L405

Leaving Transfer-Encoding unset when `resp.StatusCode == http.StatusNotModified` seems to fix that issue. 

```go
} else if resp.StatusCode == http.StatusNotModified { 
 	// for 304 responses leave Transfer-Encoding and Content-Length unset
 	resp.Header.Del("Content-Length") 
} else { 
 	...
```

I haven't checked to see if this is reproducible on the main branch (so feel free to close if this is already solved). I'll try to report back after I have a chance to check out the latest branch.

PR #775, opened 2026-04-28
fix: do not send Transfer-Encoding for 204 and 1xx status codes

babariviere:
Since 204 and 1xx responses must not have a body per their RFC, we should not set the `Transfer-Encoding: chunked` header to avoid errors from browsers / http clients.

Related to elazarl/goproxy#732


ErikPelli:
I made my own version of this and the previous pull request associated with this issue here: https://github.com/elazarl/goproxy/pull/776
Thanks for your help but I needed to clean this up and avoid multiple if statements to check every possible status code.

PR #779, opened 2026-04-30
fix(https): close client connection when target connection errors

### LINKED ISSUE #657: Sockets can be leaked if the target connection has an IO error, leading to clients hanging indefinitely
For context, I am writing this as a client connecting to a proxy built on goproxy.

When running at scale with HTTPS connections going through a proxy, I have occasionally gotten connections that hang indefinitely. To the best I can tell, this comes down to the following factors:
1. An HTTP POST request is made through HTTPS.
2. A connection is established through the proxy to the target.
3. After the connection is established, there is either a read timeout (`Error dialing to <host>: dial tcp: lookup [host] on [host-ip]: read udp [client-ip]->[host-ip]: i/o timeout`) or the host closes the socket (`Error copying to client: writeto tcp [localhost:port]->[localhost:port]: readfrom tcp [client-ip]->[host-ip]: splice: connection reset by peer`).
4. [This block gets hit](https://github.com/elazarl/goproxy/blob/master/https.go#L164) for the managing the HTTPS connection, hitting the case where only the connection to the target is closed while leaving the connection to the client open.
5. The client attempts to write the body of the POST to the socket, which hangs indefinitely as the proxy has leaked the connection.

[The block of code I linked](https://github.com/elazarl/goproxy/blob/master/https.go#L164) seems to make the following assumption:
> The other side of the connection remains open until it either times out or is reset by the client.

However, while HTTP clients commonly have *connection* and *read* timeouts, not all socket operations have timeouts associated with them, most notably writes. In this case, the request goes through Java's `HttpClient` with explicit connect and read timeouts set, which are the only timeout values exposed through the interface.

To avoid these situations, I think some solution needs to be implemented to ensure all sockets get closed at some point. For example, if it's impossible to handle asynchronously as the comment implies, could connections be placed on a list that is periodically checked for errors to close the other end of the connection?

wucm667:
Fixes #657

## Bug Description
When a target connection has an IO error (read timeout or connection reset)
during a transparent HTTPS tunnel (CONNECT Accept path), only the target-side
connection is closed. The client-side connection is left open, causing clients
to hang indefinitely waiting to write the POST body.

## Root Cause
The `copyAndClose` function in `https.go` only half-closed connections
(CloseWrite/CloseRead) on both success and error. A half-close does not
unblock a goroutine blocked on `io.Copy` reading from the same connection,
leaving `wg.Wait()` hanging and the client connection never closed.

## Fix
In `copyAndClose`, when `io.Copy` returns an error, fully `Close()` both
the destination and source connections. This unblocks any goroutine blocked
on `io.Copy` reading from those connections, allowing `wg.Wait()` to complete
and the cleanup code (proxyClientTCP.Close(), targetTCP.Close()) to run.

On success, the original graceful half-close behavior is preserved.

## Test
Added `TestTransparentTunnelClosesClientConnOnTargetError` regression test
that verifies the client connection is closed within a timeout when the target
connection is unexpectedly closed.

### DIFF
--- a/https.go
+++ b/https.go
@@ -473,12 +473,20 @@ func copyOrWarn(ctx *ProxyCtx, dst io.Writer, src io.Reader) error {
 
 func copyAndClose(ctx *ProxyCtx, dst, src halfClosable, wg *sync.WaitGroup) {
 	_, err := io.Copy(dst, src)
-	if err != nil && !errors.Is(err, net.ErrClosed) {
-		ctx.Warnf("Error copying to client: %s", err.Error())
+	if err != nil {
+		if !errors.Is(err, net.ErrClosed) {
+			ctx.Warnf("Error copying to client: %s", err.Error())
+		}
+		// Fully close dst to unblock any goroutine blocked on
+		// io.Copy reading from it. Half-close (CloseWrite/CloseRead)
+		// would not interrupt a pending read, leaving the other
+		// goroutine stuck and the client connection never closed.
+		_ = dst.Close()
+		_ = src.Close()
+	} else {
+		_ = dst.CloseWrite()
+		_ = src.CloseRead()
 	}
-
-	_ = dst.CloseWrite()
-	_ = src.CloseRead()
 	wg.Done()
 }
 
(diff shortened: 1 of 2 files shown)


ErikPelli:
@wucm667 can you please fix the 2 linting issues?

wucm667:
@ErikPelli Fixed the 2 noctx lint errors by replacing net.Listen/net.Dial with net.ListenConfig.Listen/net.Dialer.DialContext using context.Background(). Could you please check again?

ErikPelli:
Thanks!

PR #784, opened 2026-06-23
Ci/optimize GitHub actions

chrollo-lucifer-12:
Refactored the workflow file
Used stable versions for the actions
Added manual caching

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -12,19 +12,37 @@ jobs:
   build:
     name: Build And Test Go code
     runs-on: ubuntu-latest
+
+    env:
+      CGO_ENABLED: 0
+      GOFLAGS: -buildvcs=false
+
     steps:
-      - uses: actions/checkout@v6
-      
-      - uses: actions/setup-go@v6
+      - name: Checkout Repository
+        uses: actions/checkout@v4
+
+      - name: Setup Go
+        uses: actions/setup-go@v5
         with:
           go-version-file: go.mod
+          cache: false
+
+      - name: Cache go modules
+        uses: actions/cache@v4
+        with:
+          path: |
+            ~/go/pkg/mod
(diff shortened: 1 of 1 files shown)


.github/workflows/go.yml
```diff
@@ -12,30 +12,48 @@ jobs:
...
+    env:
+      CGO_ENABLED: 0
+      GOFLAGS: -buildvcs=false
+
     steps:
-      - uses: actions/checkout@v6
-      
-      - uses: actions/setup-go@v6
+      - name: Checkout Repository
+        uses: actions/checkout@v4
```
elazarl, line 22: why do we downgrade? AI slop?

.github/workflows/go.yml
```diff
@@ -12,30 +12,48 @@ jobs:
...
+
+      - name: Setup Go
+        uses: actions/setup-go@v5
         with:
           go-version-file: go.mod
+          cache: false
+
+      - name: Cache go modules
```
elazarl, line 30: why the cache?

.github/workflows/go.yml
```diff
@@ -12,30 +12,48 @@ jobs:
...
-          CGO_ENABLED: 0
+        run: go build -v ./...
 
       # Make sure to detect eventual race conditions
       # (CGO must be enabled to use -race detector)
       - name: Test
         # -count=2 ensures that test fixtures cleanup after themselves
         # because any leftover state will generally cause the second run to fail.
         run: go test -race -p 1 -v -shuffle=on -count=2 ./...
+        env:
```
ErikPelli, line 53: CGO should remain disabled, we don't use it

.github/workflows/go.yml
jeffwidman: Agreed, I thought modern setup-go handled caching automagically. I don't understand the need for this either?

PR #785, opened 2026-06-29
HTTP/2 MITM Full Support

ErikPelli:


http2.go
```diff
@@ -0,0 +1,203 @@
+package goproxy
+
+import (
+	"context"
+	"crypto/tls"
+	"io"
+	"net"
+	"net/http"
```
omriza, line 10:
```suggestion
	"net/http"
	"strings"
```

http2.go
```diff
@@ -0,0 +1,203 @@
...
+		}
+	}
+
+	// If the body was replaced by a filter, drop Content-Length so
+	// the http2 framer can stream it without a length mismatch.
+	if resp.Body != origBody {
+		w.Header().Del("Content-Length")
+	}
+
+	w.WriteHeader(resp.StatusCode)
```
omriza, line 201:
```suggestion
	// Announce pre-known trailers before WriteHeader (see handleHttp in http.go).
	announcedTrailers := len(resp.Trailer)
	if announcedTrailers > 0 {
		trailerKeys := make([]string, 0, announcedTrailers)
		for k := range resp.Trailer {
			trailerKeys = append(trailerKeys, k)
		}
		w.Header().Add("Trailer", strings.Join(trailerKeys, ", "))
	}

	w.WriteHeader(resp.StatusCode)
```

http2.go
```diff
@@ -0,0 +1,203 @@
...
+
+	if resp.Body != nil {
+		if _, err := io.Copy(w, resp.Body); err != nil {
+			ctx.Warnf("HTTP/2 MITM: error writing response body: %v", err)
+		}
+	}
+}
```
omriza, line 254:
```suggestion
	if resp.Body != nil {
		if _, err := io.Copy(w, resp.Body); err != nil {
			ctx.Warnf("HTTP/2 MITM: error writing response body: %v", err)
		}
	}

	// Forward response trailers after the body: pre-announced by name, the rest
	// (h2/gRPC send them unannounced) via http.TrailerPrefix; Flush forces
	// chunking. Mirrors handleHttp in http.go.
	if len(resp.Trailer) > 0 {
		if rc := http.NewResponseController(w); rc != nil {
			_ = rc.Flush()
		}
		if len(resp.Trailer) == announcedTrailers {
			copyHeaders(w.Header(), resp.Trailer, proxy.KeepDestinationHeaders)
		} else {
			for k, vs := range resp.Trailer {
				k = http.TrailerPrefix + k
				for _, v := range vs {
					w.Header().Add(k, v)
				}
			}
		}
	}
}
```

omriza (commented):
The h2 MITM path (`handleH2MitmStream`) forwards response headers and body but not trailers, so gRPC and other trailer-based HTTP/2 protocols break through the proxy the client gets `server closed the stream without sending trailers`. This mirrors the trailer handling `handleHttp` already does in `http.go`.

Porting the code from http.go

http2.go
```diff
@@ -0,0 +1,233 @@
+package goproxy
+
+import (
+	"context"
+	"crypto/tls"
+	"io"
```
omriza, line 7:
```suggestion
	"io"
	"mime"
```

http2.go
```diff
@@ -0,0 +1,233 @@
...
+		w.Header().Add("Trailer", strings.Join(trailerKeys, ", "))
+	}
+
+	w.WriteHeader(resp.StatusCode)
+
+	if resp.Body != nil {
+		if _, err := io.Copy(w, resp.Body); err != nil {
+			ctx.Warnf("HTTP/2 MITM: error writing response body: %v", err)
+		}
+	}
```
omriza, line 234:
```suggestion
	if resp.Body != nil {
		if shouldFlushStreaming(resp) {
			// Streaming (gRPC/Connect/SSE/chunked): flush each chunk so it reaches the
			// client as it arrives. Mirrors net/http/httputil.ReverseProxy.flushInterval.
			rc := http.NewResponseController(w)
			buf := make([]byte, 32*1024)
			for {
				nr, er := resp.Body.Read(buf)
				if nr > 0 {
					if _, ew := w.Write(buf[:nr]); ew != nil {
						ctx.Warnf("HTTP/2 MITM: error writing response body: %v", ew)
						break
					}
					_ = rc.Flush()
				}
				if er != nil {
					// Mirror net/http/httputil.ReverseProxy.copyBuffer: io.EOF is the
					// normal end of stream and context.Canceled means the client went
					// away or cancelled the request, so neither is worth logging.
					if er != io.EOF && er != context.Canceled {
						ctx.Warnf("HTTP/2 MITM: error reading response body: %v", er)
					}
					break
				}
			}
		} else {
			// Fixed-length response: let the h2 server batch writes for throughput.
			if _, err := io.Copy(w, resp.Body); err != nil {
				ctx.Warnf("HTTP/2 MITM: error writing response body: %v", err)
			}
		}
	}
```

http2.go
```diff
@@ -0,0 +1,233 @@
...
+		} else {
+			for k, vs := range resp.Trailer {
+				k = http.TrailerPrefix + k
+				for _, v := range vs {
+					w.Header().Add(k, v)
+				}
+			}
+		}
+	}
+}
```
omriza, line 254:
```suggestion
}

// shouldFlushStreaming reports whether resp should be forwarded with a flush after
// each chunk, mirroring net/http/httputil.ReverseProxy.flushInterval: Server-Sent
// Events, or any response with an unknown (-1) Content-Length (gRPC, Connect, chunked).
func shouldFlushStreaming(resp *http.Response) bool {
	if baseCT, _, _ := mime.ParseMediaType(resp.Header.Get("Content-Type")); baseCT == "text/event-stream" {
		return true
	}
	return resp.ContentLength == -1
}
```

https.go
```diff
@@ -272,13 +388,40 @@ func (proxy *ProxyHttpServer) handleHttps(w http.ResponseWriter, r *http.Request
 					}
 				}
 
+				if proxy.AllowHTTP2 {
```
omriza, line 431: `tlsConfig := defaultTLSConfig` aliases the shared global here, so when `todo.TLSConfig` is nil the `NextProtos` appends below mutate it in place: concurrent MITM handshakes race on the slice, and the appended protocols leak into every later use of the config (incl. the `defaultTLSConfig.Clone()` in the signing path). cloning first fixes both, same as the signing path already does.

```suggestion
				// Clone before appending so we never mutate the shared
				// defaultTLSConfig (or a config the TLSConfig callback reuses):
				// concurrent MITM handshakes would otherwise race on the slice,
				// and the appended protocols would leak into every later use.
				tlsConfig = tlsConfig.Clone()

				if proxy.AllowHTTP2 {
```

http2.go
```diff
@@ -0,0 +1,272 @@
...
+
+	origBody := resp.Body
+	resp = proxy.filterResponse(resp, ctx)
+	defer resp.Body.Close()
+
+	for k, vv := range resp.Header {
+		for _, v := range vv {
+			w.Header().Add(k, v)
+		}
+	}
```
omriza: small one: reuse the existing `copyHeaders` helper instead of the hand-rolled loop. tidier, and it honors `KeepDestinationHeaders` like the rest of the proxy.

```suggestion
	copyHeaders(w.Header(), resp.Header, proxy.KeepDestinationHeaders)
```

https.go
```diff
@@ -156,28 +159,103 @@ var _ halfClosable = (*net.TCPConn)(nil)
...
+				if err := ctx.Resp.Write(proxyClient); err != nil {
+					ctx.Warnf("Cannot write response that reject http CONNECT: %v", err)
+				}
+			} else {
+				// HTTP/2: write the rejection as a proper HTTP response.
+				for k, vv := range ctx.Resp.Header {
+					for _, v := range vv {
+						w.Header().Add(k, v)
+					}
+				}
```
omriza: same here, `copyHeaders` covers this and honors `KeepDestinationHeaders`.

```suggestion
				copyHeaders(w.Header(), ctx.Resp.Header, proxy.KeepDestinationHeaders)
```

http2.go
```diff
@@ -0,0 +1,268 @@
...
+		RoundTripper: parentCtx.RoundTripper,
+		certStore:    parentCtx.certStore,
+	}
+
+	req, resp := proxy.filterRequest(r, ctx)
+	if resp == nil {
+		removeH2HopByHopHeaders(req)
+		if !proxy.KeepHeader {
+			RemoveProxyHeaders(ctx, req)
+		}
```
omriza, line 161: Over h2 the `http2.Server` gives bodyless requests a non-nil empty `req.Body` with `ContentLength 0`; forwarded as-is the upstream `http2.Transport` reads that as unknown length (-1) and sends a phantom request body, which strict origins reject. mirrors what `net/http/httputil.ReverseProxy` does for `ContentLength == 0`, using `http.NoBody` so downstream filters that assume a non-nil body stay safe.

```suggestion
		if !proxy.KeepHeader {
			RemoveProxyHeaders(ctx, req)
		}

		// bodyless h2 requests arrive with a non-nil empty Body; forwarding as-is
		// makes the upstream transport send a phantom body (-1). keep it bodyless.
		if req.ContentLength == 0 {
			req.Body = http.NoBody
		}
```

ErikPelli: Can you clarify in the comment that this change is also used in httputil.ReverseProxy?

PR #787, opened 2026-07-06
fix(https): coalesce the MITM response head into a single write

### LINKED ISSUE #786: MITM response head written unbuffered, fragmenting it into dozens of tiny TLS records
**Issue**: 
In the MITM path the response head is written straight to the raw `*tls.Conn` via [`resp.Write(client)`](https://github.com/elazarl/goproxy/blob/636207d4c044d260140a1f7a6f13efb609dbadc0/https.go#L411-L419). `http.Response.Write` → `Header.Write` does a separate write per header field (and per `\r\n`), and a `*tls.Conn` emits one TLS record per `Write` — so a normal response head goes out as ~60 tiny records instead of one. 
`net/http`'s own server avoids this by wrapping the response in a `bufio.Writer`, the MITM path doesn't. 
Both the WebSocket-upgrade write (`L411`) and the normal write (`L419`) are affected.

**Impact**: 
Strict clients reject a response delivered as too many small packets. 
OpenAI Codex's WebSocket client (Rust `tungstenite`) fails the `101` upgrade with `Attack attempt detected`. its check trips at 64+ chunks averaging <128 bytes — and falls back to HTTP.

**To Reproduce**: 
stock `NewProxyHttpServer()` + `HandleConnect(AlwaysMitm)` 
CONNECT through it to any TLS upstream and count the client-facing writes for the response head. you should see ~60 chunks, all <128 bytes. 

**Version**: master (`636207d`) / v1.8.4.

shaharfreiman:
Fixes #786.

The MITM path wrote the response head straight to the raw `*tls.Conn` via `resp.Write(client)`, so `Header.Write`'s per-field writes each became their own TLS record — a normal head fragmented into ~60 tiny records. Strict clients reject that (`tungstenite`'s WebSocket handshake fails it as an "attack"), and it's wasteful regardless.

Wraps both `resp.Write(client)` sites (the WebSocket `101` head and the normal response) in a `bufio.Writer` flushed once, so the head ships as a single record — matching how `net/http`'s server writes. The body still streams (`bufio.Writer` auto-flushes at 4 KB; nothing is buffered whole).

Added `https_head_test.go`, which counts client-facing writes for a MITM'd response: **~60 without the fix, 3 with it**.

### DIFF
--- a/https_head_test.go
+++ b/https_head_test.go
@@ -72,8 +72,7 @@ func TestMitmResponseHeadIsNotFragmented(t *testing.T) {
 	proxy.OnRequest(goproxy.ReqHostIs(upstream.Listener.Addr().String())).HandleConnect(goproxy.AlwaysMitm)
 
 	var writes atomic.Int32
-	lc := net.ListenConfig{}
-	ln, err := lc.Listen(context.Background(), "tcp", "127.0.0.1:0")
+	ln, err := (&net.ListenConfig{}).Listen(context.Background(), "tcp", "127.0.0.1:0")
 	require.NoError(t, err)
 	server := &http.Server{Handler: proxy, ReadHeaderTimeout: 10 * time.Second}
 	go func() {
--- a/https_head_test.go
+++ b/https_head_test.go
@@ -1,0 +1,104 @@
+package goproxy_test
+
+import (
+	"context"
+	"crypto/tls"
+	"io"
+	"net"
+	"net/http"
+	"net/http/httptest"
+	"net/url"
+	"sync/atomic"
+	"testing"
+	"time"
+
+	"github.com/elazarl/goproxy"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+type countingConn struct {
+	net.Conn
+	writes *atomic.Int32
+}
+
+func (c *countingConn) Write(p []byte) (int, error) {
+	c.writes.Add(1)
+	return c.Conn.Write(p)
+}
+
+type countingListener struct {
+	net.Listener
+	writes *atomic.Int32
+}
+
+func (l *countingListener) Accept() (net.Conn, error) {
+	c, err := l.Listener.Accept()
+	if err != nil {
+		return c, err
+	}
+	return &countingConn{Conn: c, writes: l.writes}, nil
+}
+
+// TestMitmResponseHeadIsNotFragmented ensures the MITM response head (status line
+// + headers + terminator) is written to the client as a single buffered flush
+// rather than one tiny TLS record per header field. Fragmenting it into dozens of
+// tiny records breaks strict clients (e.g. tungstenite's WebSocket handshake
+// rejects a response delivered as too many small packets).
+func TestMitmResponseHeadIsNotFragmented(t *testing.T) {
+	upstream := httptest.NewTLSServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		for _, kv := range [][]string{
+			{"Cache-Control", "no-cache"},
+			{"Content-Language", "en"},
+			{"Referrer-Policy", "strict-origin-when-cross-origin"},
+			{"Strict-Transport-Security", "max-age=31536000; includeSubDomains"},
+			{"Vary", "Accept-Encoding"},
+			{"X-Content-Type-Options", "nosniff"},
+			{"X-Frame-Options", "DENY"},
+			{"X-One", "1"},
+			{"X-Two", "2"},
+			{"X-Three", "3"},
+			{"X-Four", "4"},
+			{"X-Five", "5"},
+		} {
+			w.Header().Set(kv[0], kv[1])
+		}
+		_, _ = io.WriteString(w, "ok")
+	}))
+	defer upstream.Close()
+
+	proxy := goproxy.NewProxyHttpServer()
+	proxy.Tr.TLSClientConfig = &tls.Config{InsecureSkipVerify: true}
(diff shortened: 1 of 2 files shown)


https_head_test.go
```diff
@@ -0,0 +1,93 @@
...
+	"testing"
+
+	"github.com/elazarl/goproxy"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+type countingConn struct {
+	net.Conn
+	writes *int32
```
ErikPelli: Instead of int32 with atomic operations, directly use an atomic.Int32 type (https://pkg.go.dev/sync/atomic#Int32).
Or atomic.Uint32/Uint64/Int64, what you think is better.

https_head_test.go
```diff
@@ -0,0 +1,93 @@
...
+
+// TestMitmResponseHeadIsNotFragmented ensures the MITM response head (status line
+// + headers + terminator) is written to the client as a single buffered flush
+// rather than one tiny TLS record per header field. Fragmenting it into dozens of
+// tiny records breaks strict clients (e.g. tungstenite's WebSocket handshake
+// rejects a response delivered as too many small packets).
+func TestMitmResponseHeadIsNotFragmented(t *testing.T) {
+	upstream := httptest.NewTLSServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		h := w.Header()
+		for _, kv := range [][2]string{
```
ErikPelli: Directly use [][]string with one header for each line, it would be much more readable

https_head_test.go
```diff
@@ -0,0 +1,93 @@
...
+	proxy.OnRequest(goproxy.ReqHostIs(upstream.Listener.Addr().String())).HandleConnect(goproxy.AlwaysMitm)
+
+	var writes int32
+	ln, err := net.Listen("tcp", "127.0.0.1:0")
+	require.NoError(t, err)
+	server := &http.Server{Handler: proxy}
+	go func() { _ = server.Serve(&countingListener{Listener: ln, writes: &writes}) }()
+	defer server.Close()
+
+	proxyURL, _ := url.Parse("http://" + ln.Addr().String())
```
ErikPelli: I don't like this url.Parse, you can directly build an url.URL using:
url.URL{
 Scheme: "http",
 Host: ln.Addr().String()
}

https_head_test.go
```diff
@@ -0,0 +1,93 @@
...
+	client := &http.Client{Transport: &http.Transport{
+		Proxy:           http.ProxyURL(proxyURL),
+		TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
+	}}
+
+	resp, err := client.Get(upstream.URL + "/")
```
ErikPelli: Don't use .Get(0, or else the linter will complain, see the other tests.
Use NewRequestWithContext() and then client.Do().

ErikPelli (changes requested):

https_head_test.go
```diff
@@ -0,0 +1,93 @@
...
+	require.NoError(t, err)
+	_, _ = io.Copy(io.Discard, resp.Body)
```
ErikPelli: I think the discard is automatically done when you close the body in the line later, no need to do this manually

https_head_test.go
```diff
@@ -0,0 +1,93 @@
...
+
+	proxy := goproxy.NewProxyHttpServer()
+	proxy.Tr.TLSClientConfig = &tls.Config{InsecureSkipVerify: true}
+	proxy.OnRequest(goproxy.ReqHostIs(upstream.Listener.Addr().String())).HandleConnect(goproxy.AlwaysMitm)
+
+	var writes int32
+	ln, err := net.Listen("tcp", "127.0.0.1:0")
+	require.NoError(t, err)
+	server := &http.Server{Handler: proxy}
+	go func() { _ = server.Serve(&countingListener{Listener: ln, writes: &writes}) }()
```
ErikPelli: Put the` _ = server.Serve(...)` inside a new line, no need to make everything in one huge line:

go func() {
   _ = server.Serve(...)
}()

ErikPelli:
@shaharfreiman can you fix the linter issue?
```
  Error: https_head_test.go:75:23: net.Listen must not be called. use (*net.ListenConfig).Listen (noctx)
```

The function it suggests uses the context

https_head_test.go
shaharfreiman: Used atomic.Int32

https_head_test.go
shaharfreiman: I agree. done

https_head_test.go
shaharfreiman: Checked and you are right. Removed

https_head_test.go
```diff
@@ -0,0 +1,104 @@
...
+	}))
+	defer upstream.Close()
+
+	proxy := goproxy.NewProxyHttpServer()
+	proxy.Tr.TLSClientConfig = &tls.Config{InsecureSkipVerify: true}
+	proxy.OnRequest(goproxy.ReqHostIs(upstream.Listener.Addr().String())).HandleConnect(goproxy.AlwaysMitm)
+
+	var writes atomic.Int32
+	lc := net.ListenConfig{}
+	ln, err := lc.Listen(context.Background(), "tcp", "127.0.0.1:0")
```
ErikPelli:
```suggestion
	ln, err := (&net.ListenConfig{}).Listen(context.Background(), "tcp", "127.0.0.1:0")
```

shaharfreiman:
> @shaharfreiman can you fix the linter issue?
> 
> ```
>   Error: https_head_test.go:75:23: net.Listen must not be called. use (*net.ListenConfig).Listen (noctx)
> ```
> 
> The function it suggests uses the context

Fixed. Thanks

PR #793, opened 2026-07-22
Fix unexpected scheme edge case in MITM

### LINKED ISSUE #792: Go 1.26: MITM fails for absolute-form requests with a different scheme
Hey, while testing `goproxy v1.8.5` with Go 1.26, I found a compatibility issue in the MITM request path.

## Issue

When goproxy detects a TLS connection inside a CONNECT tunnel, it uses `https` as the request scheme. Some clients, including npm in certain proxy flows, may send an absolute-form request target using `http`:

```text
http://registry.npmjs.org:443/-/npm/v1/security/advisories/bulk
```

The current MITM code checks whether the request URL starts with the detected scheme:

https://github.com/elazarl/goproxy/blob/v1.8.5/https.go#L306-L307

Because the request starts with `http://` rather than `https://`, it is treated like a relative URL and concatenated into:

```text
https://registry.npmjs.org:443http://registry.npmjs.org:443/-/npm/v1/security/advisories/bulk
```

Go 1.26 rejects this malformed authority, returning a nil URL and an `invalid port ":443http:" after host` error. goproxy retains that error and later aborts the request:

https://github.com/elazarl/goproxy/blob/v1.8.5/https.go#L353-L359

The client receives an EOF.

## Impact

Absolute-form requests whose scheme differs from the scheme detected by the MITM path fail when goproxy is built with Go 1.26. I reproduced this on Linux and macOS with Go 1.26.5. The current master branch appears to contain the same reconstruction logic.

## To Reproduce

```go
package main

import (
	"fmt"
	"net/url"
)

func main() {
	scheme := "https"
	connectHost := "registry.npmjs.org:443"
	requestURL, _ := url.Parse("http://registry.npmjs.org:443/-/npm/v1/security/advisories/bulk")

	reconstructed := scheme + "://" + connectHost + requestURL.String()
	parsed, err := url.Parse(reconstructed)

	fmt.Printf("URL: %v\\nerror: %v\\n", parsed, err)
}
```

With Go 1.26, this returns:

```text
URL: <nil>
error: parse "https://registry.npmjs.org:443http://registry.npmjs.org:443/-/npm/v1/security/advisories/bulk": invalid port ":443http:" after host
```

## Expected Behavior

goproxy should recognize that the request URL is already absolute even when its scheme differs from the detected MITM scheme, and should not concatenate the CONNECT host and scheme onto it.

ErikPelli:
Fixes https://github.com/elazarl/goproxy/issues/792

### DIFF
--- a/https.go
+++ b/https.go
@@ -303,8 +303,12 @@ func (proxy *ProxyHttpServer) handleHttps(w http.ResponseWriter, r *http.Request
 				req.RemoteAddr = r.RemoteAddr
 				ctx.Logf("req %v", r.Host)
 
-				if !strings.HasPrefix(req.URL.String(), scheme+"://") {
+				if !req.URL.IsAbs() {
+					// Origin-form request target (/path)
 					req.URL, err = url.Parse(scheme + "://" + r.Host + req.URL.String())
+				} else {
+					// Absolute-form request target
+					req.URL.Scheme = scheme
 				}
 
 				if continueLoop := func(req *http.Request) bool {


Sahilb315:
Hey @ErikPelli, thanks again for the fix. Do you have a rough timeline for when this might be merged and included in a release?

ErikPelli:
@Sahilb315 I'm going to merge it now and release a new version next week

PR #794, opened 2026-08-04
Stream MITM response bodies

krissetto:
Fixes SSE streaming through HTTPS MITM.  

Since v1.8.5/#787, the existing flush runs only after `resp.Write` finishes, but `resp.Write` includes the long-lived response body and therefore never returns for SSE. Small events remain stuck in the 4 KiB buffer.  

This change buffers only the response headers, then streams the body writes directly, preserving header coalescing.  
Adds regression coverage.

### DIFF
--- a/https.go
+++ b/https.go
@@ -2,6 +2,7 @@ package goproxy
 
 import (
 	"bufio"
+	"bytes"
 	"context"
 	"crypto/tls"
 	"errors"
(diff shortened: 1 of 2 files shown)


ErikPelli:
Sorry to hear that #787 caused this issue to you, I verified that your solution perfectly works without removing the good change that the previous PR added, so I'm going to merge it and release a new version.
Thanks!

krissetto:
Thanks for the quick review and merge!

PR #797, opened 2026-08-28
fix(mitm): never frame bodiless responses as chunked

doringeman:
Avoid adding chunked transfer encoding to MITM responses that cannot contain a message body.

The existing check relies on `resp.Body == http.NoBody`. Empty HTTP/2 responses use a different body implementation,
causing HEAD, 204, and 304 responses to be framed as chunked when they are written back to the HTTP/1.1 client.

### DIFF
--- a/https.go
+++ b/https.go
@@ -520,7 +520,7 @@ func (proxy *ProxyHttpServer) handleHttps(w http.ResponseWriter, r *http.Request
 					resp = proxy.filterResponse(resp, ctx)
 					bodyModified := resp.Body != origBody
 					defer resp.Body.Close()
-					if resp.Body != http.NoBody && (bodyModified ||
+					if responseBodyAllowed(req, resp) && resp.Body != http.NoBody && (bodyModified ||
 						(resp.ContentLength <= 0 && resp.Header.Get("Content-Length") == "")) {
 						// Return chunked encoded response when we don't know the length of the resp, if the body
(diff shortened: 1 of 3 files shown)


doringeman:
@ErikPelli @krissetto PTAL 🙌

ErikPelli:
Thanks for the pull request, can you open a detailed issue with the reason behind this change? Is http/2 currently broken?

doringeman:
Hey @ErikPelli, of course, here it is: https://github.com/elazarl/goproxy/issues/800.

mark-bradshaw:
Thanks for this, we carry an equivalent local patch. One thing to check: the helper takes `req`, but inside the MITM loop `req` can be nil when a request handler returned a response directly (`filterRequest` returns `nil, resp`). It would panic there. `ctx.Req` is always set at that point and is the safer input.

Also worth a test: a response handler that replaces the body of a 304 or 204 over an HTTP/1 origin hits the same chunked path, since the replacement body is not `http.NoBody`. The status-and-method check covers it.

doringeman:
Thanks @mark-bradshaw, addressed in https://github.com/elazarl/goproxy/pull/797/commits/a154b4ed624f400b8a808b91ba0e7c10a6ef21c2.

ErikPelli:
Thanks for the important discovery, it's now publicly released!

PR #803, opened 2026-09-04
Refactor tests: testify, shared helpers, run ext and examples tests in CI

### LINKED ISSUE #581: Refactor tests
Current tests are legacy and really old, no one cared much about them.
Failing tests are not reporting the errors to the pipeline result, the whole pipeline should fail, and not report a success.

1. Try to fix the commented tests (with //), and re-enable them
2. Rewrite all the tests using testify library (currently used for http1parser_test package)
3. Refactor the code of the tests to make them more readable

mark-bradshaw:
Closes #581

## Problem

The test suite was inconsistent and partly broken. Each package had its own copy of the same proxy helpers. Assertions were a mix of hand written `if` checks and testify. CI only ran the root module, so the tests in `ext/` and `examples/` never ran. The `ext/image` tests have been failing for a long time and nobody could see it.

## Solution

- Every test file now uses testify (`assert` and `require`).
- Shared helpers live in one new package, `internal/testutil`, as requested in the issue. The duplicated copies in each package are gone.
- `ext/go.mod` and `examples/go.mod` now point at the local checkout, so their tests run against the current code and can use the shared helpers. This does not affect people who import goproxy.
- CI now runs the `ext/` and `examples/` test suites too.
- Tests were renamed where the old name did not say what the test checks.

## Bugs found and fixed along the way

- `TestNoProxyHeadersHttps` was not testing anything. It ignored a TLS error, so the backend handler never ran. It now asserts on the real response.
- `ext/image` tests read fixtures from the wrong path and always failed. Fixed.
- Three tests used hard coded ports (5000, 8082, 10234). Port 5000 is taken on every recent Mac, so the suite failed locally. They now use ephemeral ports.
- The concurrency limiter test leaked goroutines that waited forever. The requests now carry cancellable contexts, and the test also confirms a blocked request proceeds once a slot frees up.
- Debug prints (`println`, `log.Printf`) were removed from tests.

## The one commented out test

`TestIcyResponse` was the only test that was commented out. It expects the proxy to forward a raw `ICY 200 OK` status line. Go's HTTP transport rejects that line and the proxy returns a 500, so making it pass needs a change to the proxy itself, not the test. I removed the dead block rather than leave it. Happy to open a separate issue for ICY support if you want it.

## About the size

You mention in the docs about not making large PRs.  As a long time dev, I get it, but the issue specifically asked for this work and since it's in test code I figured an exception could be made.  If you want it broken up I can do that, but it might be nice to just get this all done in one feel swoop.

This PR is large: about 1100 lines added and 1650 removed across 20 files. It is almost entirely test code. Outside of tests, the only changes are the CI workflow and the two `go.mod`/`go.sum` pairs. No proxy source code changed.

To check that nothing was lost: 97 top level tests before, 96 after. The one removed is `TestIcyResponse`, which was already empty. All other removed names are renames within the same file. Subtest count is unchanged.

## Proof

```
go test -race -p 1 -shuffle=on -count=2 ./...   # root, ext, examples: all ok
golangci-lint run ./...                           # 0 issues
```

Each rewritten test was also checked by hand with a small mutation step: change one expected value or one input so the test *should* fail, run it, confirm it fails with a sensible message, then put it back. Examples: comparing the panda image against the football image failed with a pixel difference; raising the concurrency limit from 1 to 2 made the "more than the limit" case fail with "limiter was too fast"; corrupting three expected strings in the jquery test produced three failures. A test that has never been seen to fail is not proof of anything, so this step was done for every file that was rewritten.

## Please look closely at

- `proxy_test.go`: the biggest file, and where the port and helper changes are concentrated.
- `ext/limitation/concurrency_test.go`: fully restructured, not a line by line conversion.
- `internal/http1parser/header_test.go`: `require.Error` became `require.ErrorIs(t, err, io.EOF)` after confirming that is the error actually returned.
- `.github/workflows/go.yml`: the new test steps for `ext/` and `examples/`.


ErikPelli:
Can you rebase this branch over main and resolve the conflicts? Also, a few tests have been added, maybe they need a refactor too.
Thanks for the pull request!

mark-bradshaw:
Rebased on master and resolved the conflicts. They were all in `proxy_test.go` and only cosmetic: upstream removed some redundant parens on lines this PR had already rewritten.

Also refactored the tests that landed in the meantime:

- `connect_reply_internal_test.go` — hand-written `if got != want` checks now use `assert`.
- `https_bodiless_test.go` — was building its own proxy server and client; now uses the shared `testutil.NewProxy` helper.
- `TestMitmConnectNormalizesDefaultPortInURL` — used `oneShotProxy`, which this PR removes, so it now uses `testutil.NewProxy` too.

All three modules pass with the CI flags (`go test -race -p 1 -shuffle=on -count=2 ./...`) on Go 1.24.0, and also on 1.26.5 and 1.27.1. `golangci-lint run ./...` reports 0 issues.

One thing I noticed but left alone: `examples/go.mod` replaces `github.com/elazarl/goproxy => ../` but not `github.com/elazarl/goproxy/ext => ../ext`, so the new examples test step runs against a published `ext` rather than the local one. Happy to add the replace here or in a separate PR, whichever you prefer.

PR #809, opened 2026-09-22
Fix: prevent panic on malformed URL during MITM

### LINKED ISSUE #808: Panic: nil pointer dereference in https.go when url.Parse fails during MITM
### Description
In version `v1.9.1`, a regression was introduced in `https.go` during MITM URL reconstruction. If `url.Parse` fails (e.g., due to a malformed inner request containing invalid characters), the code attempts to log the error but triggers a nil pointer dereference panic, crashing the entire proxy process.

### Root Cause
The issue is located in the MITM request handling loop in `https.go`. The code performs the following sequence:

```go
req.URL, err = url.Parse(scheme + "://" + hostToUse + req.URL.String())
if err != nil {
    ctx.Warnf("Cannot parse URL %s: %v", scheme+"://"+hostToUse+req.URL.String(), err) // PANIC HERE
    return
}
```

When `url.Parse` encounters an invalid URL, it returns an error and a `nil` pointer for the `*url.URL` object. The multiple assignment overwrites `req.URL` with `nil`. Subsequently, calling `req.URL.String()` inside the `if err != nil` block causes a `panic: runtime error: invalid memory address or nil pointer dereference`.

### Example
The panic is triggered by malformed requests (likely from tools like Nuclei) attempting SQL Injection via Host header/path manipulation. Example of URL that trigger the parse failure: `http://[::1]' UNION SELECT '//etc/passwd` (Fails with: `invalid port "' UNION SELECT '" after host`)

### Proposed Fix
Store the constructed URL string in a local variable *before* calling `url.Parse`. Use this variable in the error log to avoid dereferencing a potentially nil `req.URL`.
```go
urlToParse := scheme + "://" + hostToUse + req.URL.String()
req.URL, err = url.Parse(urlToParse)
if err != nil {
    ctx.Warnf("Cannot parse URL %s: %v", urlToParse, err)
    return
}
```
This fix prevents the proxy from crashing when encountering intentionally malformed requests, allowing it to safely log the warning.

khotiashova:
### Summary
This PR fixes a critical regression introduced in `v1.9.1` where the proxy crashes with a `nil pointer dereference` panic when encountering malformed HTTP requests inside a MITM tunnel. 

### Testing
Added a regression test `TestMitmMalformedRequestNoPanic`  to emulate a malformed request inside a MITM tunnel, ensuring the proxy handles it gracefully without panicking.

For more details, see issue
Closes #808


https.go
```diff
@@ -493,9 +493,10 @@ func (proxy *ProxyHttpServer) handleHttps(w http.ResponseWriter, r *http.Request
 						hostToUse = r.Host // Fallback, if the internal Host header is missing
 					}
 
-					req.URL, err = url.Parse(scheme + "://" + hostToUse + req.URL.String())
+					urlToParse := scheme + "://" + hostToUse + req.URL.String()
+					req.URL, err = url.Parse(urlToParse)
```
WGH-, line 497: You leave `req.URL` as `nil` after a failure. Although the `return` after `ctx.Warnf` discards `http.Request`, and nobody should see it (I think?), potentially the code might be modified later to accept such semi-broken requests. I think we should conservatively keep the original URL as is in case of failure.

khotiashova: You are absolutely right. Thank you for pointing out this side effect. 
I've updated the code to parse the URL into a temporary variable first. 
Could you please take another look?

WGH- (approved):

